PR #3, opened 2021-02-15
[CI] Use GitHub instead of Travis.

### LINKED ISSUE #2: Travis CI is no longer supported
Travis CI has changed how open source projects are supported, and Hyperledger can no longer build there. I suggest GitHub Actions as a replacement.

ggwpez:
Replace the Travis configuration with GitHub actions.  
The codeclimate part is left out since it is currently not setup for this repo.  

Closes #2


ggwpez:
Draft until we merged everything in gitlab.

.scripts/check-notice-authors.sh
```diff
@@ -25,17 +25,19 @@ exit_code=0
 # Call with an ancestor whereas all commits newer than the ancestor are checked.
 base="$1"
 if [ -z "$base" ]; then
-    commits="$(git rev-list --reverse HEAD)"
+    commits="$(git rev-list --no-merges --reverse HEAD)"
```
matthiasgeihs, line 28: This change can probably be reverted since the most recent CI script version.

.scripts/check-notice-authors.sh
```diff
@@ -25,17 +25,19 @@ exit_code=0
...
 else
-    commits="$(git rev-list --reverse $base..HEAD)"
+    commits="$(git rev-list --no-merges --reverse $base..HEAD)"
```
matthiasgeihs, line 30: Same as comment above.

.scripts/check-notice-authors.sh
ggwpez: Yes since we do not need to check the DCO message for merge commits.  
The case that we have merge commits an a merge request is rare, but could happen.

PR #4, opened 2021-02-28
Cache first channel update

matthiasgeihs:
When interacting on a channel directly after channel opening, it can happen that the other side has not yet completed the channel opening procedure. In this case, there was a notice about a channel update for an unknown channel and the update would never go through. This change fixes this bug by caching the first channel update until the channel opening is complete.

### DIFF
--- a/client/client_role_test.go
+++ b/client/client_role_test.go
@@ -38,11 +38,21 @@ func NewSetups(rng *rand.Rand, names []string) []ctest.RoleSetup {
 
 	for i := 0; i < n; i++ {
 		acc := wtest.NewRandomAccount(rng)
+
+		// The use of a delayed funder simulates that channel participants may
+		// receive their funding confirmation at different times.
+		var funder channel.Funder
+		if i == 0 {
+			funder = &logFunderWithDelay{log.WithField("role", names[i])}
+		} else {
+			funder = &logFunder{log.WithField("role", names[i])}
+		}
+
 		setup[i] = ctest.RoleSetup{
 			Name:        names[i],
 			Identity:    acc,
 			Bus:         bus,
-			Funder:      &logFunder{log.WithField("role", names[i])},
+			Funder:      funder,
 			Adjudicator: &logAdjudicator{log.WithField("role", names[i]), sync.RWMutex{}, nil},
 			Wallet:      wtest.NewWallet(),
 			Timeout:     roleOperationTimeout,
@@ -57,6 +67,10 @@ type (
 		log log.Logger
 	}
 
+	logFunderWithDelay struct {
+		log log.Logger
+	}
+
 	logAdjudicator struct {
 		log         log.Logger
 		mu          sync.RWMutex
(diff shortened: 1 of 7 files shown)


client/test/alice.go
```diff
@@ -46,8 +46,6 @@ func (r *Alice) Execute(cfg ExecConfig) {
 func (r *Alice) exec(_cfg ExecConfig, ch *paymentChannel) {
 	cfg := _cfg.(*AliceBobExecConfig)
 	we, them := r.Idxs(cfg.Peers())
-	// 1st stage - channel controller set up
-	r.waitStage()
```
ggwpez, line 50: Line 38 probably needs to be changed from `4` to `3`.

client/test/bob.go
```diff
@@ -38,9 +38,6 @@ func (r *Bob) exec(_cfg ExecConfig, ch *paymentChannel, propHandler *acceptNextP
 	cfg := _cfg.(*AliceBobExecConfig)
 	we, them := r.Idxs(cfg.Peers())
 
-	// 1st stage - channel controller set up
-	r.waitStage()
```
ggwpez, line 42: Same as with `Alice` in line 29.

client/proposal.go
```diff
@@ -549,6 +557,40 @@ func enableVer0Cache(ctx context.Context, c wire.Cacher) {
...
+}
+
+func (c *Client) releaseVer1Cache() {
+	c.log.Trace("Releasing version 1 cache")
+
+	c.version1Cache.mu.Lock()
+	defer c.version1Cache.mu.Unlock()
+
+	c.version1Cache.enabled--
+	for _, u := range c.version1Cache.cache {
```
ggwpez, line 576: I thought about it again and i think this is still incorrect when more than one channel is opened.  
Since it releases *all* buffered messages. Imagine:  
```go
enableVer1Cache()
// Version 1 Message is received and buffered
enableVer1Cache()
// Version 1 Message is received and buffered
releaseVer1Cache()
// Now BOTH Messages are replayed instead of just the first.
```

client/client_role_test.go
```diff
@@ -38,11 +38,21 @@ func NewSetups(rng *rand.Rand, names []string) []ctest.RoleSetup {
 
 	for i := 0; i < n; i++ {
 		acc := wtest.NewRandomAccount(rng)
+
+		// The use of a delayed funder simulates that channel participants may
+		// receive their funding confirmation at different times.
+		var funder channel.Funder
+		if i == 0 {
+			funder = &logFunderWithDelay{log.WithField("role", names[i])}
```
ggwpez, line 46: Good idea :+1:

ggwpez (changes requested):

client/test/alice.go
matthiasgeihs: That's right, good catch!

client/proposal.go
ggwpez: Ok i see, this works since you start the handler in a go-routine, otherwise the mutex would already be locked.

ggwpez (approved):
Looks good! Just needs a rebase

PR #10, opened 2021-03-04
Add PeerNotRespondingError

manoranjith:
This error identifies the scenario where the context expires when waiting for a response for a channel update or a channel proposal from the peer.

This was the best place I was able to detect this case. But I am not sure if context expiring could always be treated as "Peer Failing To Respond". Request your comments on this.


client/update.go
```diff
@@ -411,3 +419,8 @@ func makeChannelUpdate(next *channel.State, actor channel.Index) ChannelUpdate {
 		ActorIdx: actor,
 	}
 }
+
+// Error implements error interface for PeerNotRespondingError.
+func (e PeerNotRespondingError) Error() string {
```
ggwpez, line 424: Could you add constructors for both error types which also accept an error and use the `WithStack` function?  
Then the context error would be passed on instead of being discarded (like now). You can take a look at `channel/errors.go`.

client/update.go
```diff
@@ -411,3 +419,8 @@ func makeChannelUpdate(next *channel.State, actor channel.Index) ChannelUpdate {
 		ActorIdx: actor,
 	}
 }
+
```
ggwpez, line 422: I still think it is sad that we now mix error checking functions like `Is…Error` and `errors.Is` but OK… its too late now.

client/update.go
```diff
@@ -185,6 +189,10 @@ func (c *Channel) update(ctx context.Context, next *channel.State) (err error) {
 
 	pidx, res, err := resRecv.Next(ctx)
 	if err != nil {
+		// Context expiry here implies we have timedout waiting for response from the peer.
+		if pcontext.IsContextError(err) {
+			return errors.WithStack(PeerNotRespondingError("peer did not respond to the channel proposal"))
```
ggwpez, line 194: This function still uses `err` in a defer in line 161, better set it to the newly created error here.

client/update.go
```diff
@@ -185,6 +189,10 @@ func (c *Channel) update(ctx context.Context, next *channel.State) (err error) {
 
 	pidx, res, err := resRecv.Next(ctx)
 	if err != nil {
+		// Context expiry here implies we have timedout waiting for response from the peer.
```
ggwpez, line 192: Please add a comment to `Update` in L133 about the new error.

client/proposal.go
```diff
@@ -264,6 +265,10 @@ func (c *Client) proposeTwoPartyChannel(
 
 	env, err := receiver.Next(ctx)
 	if err != nil {
+		// Context expiry here implies we have timedout waiting for response from the peer.
```
ggwpez, line 268: Pleas add a comment to `ProposeChannel` in L132 about the error.

ggwpez, line 268:
```suggestion
```
I think the code is easy enough to understand here.

client/update.go
ggwpez, line 192:
```suggestion
```

client/update.go
```diff
@@ -93,6 +94,9 @@ type (
 		req     *msgChannelUpdate
 		called  atomic.Bool
 	}
+
+	// PeerNotRespondingError indicates that a peer has not responded within the expected time period.
+	PeerNotRespondingError string
```
ggwpez, line 99: Im not quite happy with the name, since it sound like the peer is not responding at all.  
But in this case it only did not respond to a specific request (or we did not receive it).  
Maybe something like `RequestTimedOutError` or `NoResponseError`.

ggwpez (changes requested):
> But I am not sure if context expiring could always be treated as "Peer Failing To Respond".

When the user closes the Channel, Client or Context while waiting for a request this error would also be triggered, but i think this is ok.  
  
Regarding testing; this is untested even before you modified it :man_facepalming: Maybe i find a place to test it.

client/update.go
manoranjith: I think `RequestTimedOutError` looks better. I will use this one.

client/update.go
manoranjith: I have used `errors.As` in the tests for the newly added error. Should I instead add an `Is..Error` function that internal uses `errors.Cause` as in other places in the project ? @ggwpez

client/update.go
manoranjith: I have added `newRequestTimedOutError` function.

manoranjith:
@ggwpez I have addresses most of the review comments and also rebased onto the tip of dev branch. I have left the two unaddressed comments in the unresolved state.

client/update.go
matthiasgeihs: It would be nice if it was consistent across the framework. On the other hand, I don't like introducing Is...Error functions all over the place, as this is pretty much redundant to the errors.Is functionality. Or am I wrong on this? Any other views and suggestions on this are appreciated.

client/proposal.go
```diff
@@ -264,6 +266,9 @@ func (c *Client) proposeTwoPartyChannel(
 
 	env, err := receiver.Next(ctx)
 	if err != nil {
+		if pcontext.IsContextError(err) {
+			return nil, newRequestTimedOutError("channel proposal", err.Error())
```
matthiasgeihs, line 270: Wouldn't it be better to wrap the error object instead of just the error string?

matthiasgeihs (commented):
Please see my comment on the custom error object. I was surprised that it does not wrap the underlying error, but just the error string.

client/proposal.go
manoranjith: I wrap the `newRequestTimedOut` error instead of the actual error because it is this error that the user of the API will be checking for. If we do it otherwise: wrap the actual error and add the `newRequestTimedOutError` as a message, I think it would not be possible for to check for `RequestTimedOutError`.

manoranjith: I add the error string, so that the details of the original error message is still retained.

client/update.go
manoranjith: I agree with you @matthiasgeihs. Personally I also prefer to use `errors.Is` over `Is..Error` function.

client/update_test.go
```diff
@@ -42,3 +43,19 @@ func TestChannel_Update_NilArgs(t *testing.T) {
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "context")
 }
+
+func TestRequestTimedOutError(t *testing.T) {
+	var err error = errors.WithStack(newRequestTimedOutError("", ""))
```
ggwpez, line 48: It should already have a stacktrace now.

client/update.go
ggwpez: Ok then lets leave it as is.

client/update.go
```diff
@@ -185,6 +191,10 @@ func (c *Channel) update(ctx context.Context, next *channel.State) (err error) {
 
 	pidx, res, err := resRecv.Next(ctx)
 	if err != nil {
+		if pcontext.IsContextError(err) {
+			err := newRequestTimedOutError("channel update", err.Error())
```
ggwpez, line 195:
```suggestion
			err = newRequestTimedOutError("channel update", err.Error())
```

client/update.go
```diff
@@ -411,3 +422,12 @@ func makeChannelUpdate(next *channel.State, actor channel.Index) ChannelUpdate {
 		ActorIdx: actor,
 	}
 }
+
+// Error implements error interface for PeerNotRespondingError.
```
ggwpez, line 426:
```suggestion
// Error implements error interface for RequestTimedOutError
.
```

manoranjith: Updated.

client/update.go
manoranjith: Updated.

client/update_test.go
manoranjith: Yeah, i missed to remove `WithStack` after adding the `newRequestTimedOutError` in previous review iteration. It have updated it.

manoranjith:
I have addressed all your review comments (3) in the 2nd iteration. @ggwpez

client/proposal.go
manoranjith: @matthiasgeihs However, if the underlying error is to be retained along with this error. Then one option is to store the actual error in the `RequestTimeOutError` struct and write an `Unwrap` method for it (see [here](https://play.golang.org/p/iJKrTEQApuq)) which will return the underlying error.

ggwpez (approved):
Im fine with this.

PR #11, opened 2021-03-04
Extend PeerRejectedProposalError for rejected updates

manoranjith:
I have renamed `PeerRejectedProposalError` to `PeerRejectedError`, so that the same error type can be used when peer rejects an update as well.

The other option could be to use two error types. Request your feedback on this.

### DIFF
--- a/client/proposal.go
+++ b/client/proposal.go
@@ -58,10 +58,11 @@ type (
 		called atomic.Bool
 	}
 
-	// PeerRejectedProposalError indicates the channel proposal has been
+	// PeerRejectedError indicates the channel proposal or channel update was
 	// rejected by the peer.
-	PeerRejectedProposalError struct {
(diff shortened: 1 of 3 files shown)


manoranjith:
I found the CI pipeline is failing due to race tests. When I tried on my local machine it fails both on this branch and also on `dev`. Where you also facing this with any of the previous PRs ?

manoranjith:
The CI pipeline seems to fail because of failing race tests. Interestingly, in the other PRs (#10) it seems to pass. But on my machine it seems to fail on `dev` as well as branches corresponding to both this and the other PR. Did you encounter race tests failing occasionally in  any of the other PRs ? @matthiasgeihs @ggwpez

manoranjith:
I have addressed all your review comments in the 2nd iteration. @ggwpez

ggwpez (approved):

PR #17, opened 2021-03-09
🐛 [backend/ethereum/wallet/simple] Unexported and protected accounts

### LINKED ISSUE #16: `Accounts` in simple.Wallet should be unexported and protected against concurrent access
The simple.Wallet implementation holds a struct `Accounts`. This is currently exported and not protected against concurrent access.

We should unexport it (i.e., `Accounts` -> `accounts`) and protect it with a read-write mutex.

cryptphil:
Closes #16

### DIFF
--- a/backend/ethereum/wallet/simple/wallet.go
+++ b/backend/ethereum/wallet/simple/wallet.go
@@ -69,18 +69,20 @@ func (w *Wallet) NewRandomAccount(prng *rand.Rand) wallet.Account {
 	if err != nil {
 		log.Panicf("Creating account: %v", err)
 	}
+
 	w.mutex.Lock()
 	defer w.mutex.Unlock()
 
 	addr := crypto.PubkeyToAddress(privateKey.PublicKey)
-	if !w.contains(addr) {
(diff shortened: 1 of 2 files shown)


backend/ethereum/wallet/simple/wallet.go
```diff
@@ -59,17 +69,19 @@ func (w *Wallet) NewRandomAccount(prng *rand.Rand) wallet.Account {
 	if err != nil {
 		log.Panicf("Creating account: %v", err)
 	}
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
 
 	addr := crypto.PubkeyToAddress(privateKey.PublicKey)
-	if !w.Contains(addr) {
+	if !w.contains(addr) {
```
matthiasgeihs, line 76: If your prng is fine, this should never happen.

The current behavior is not plausible to me. If you want to make sure that a new account is created, you should either retry key generation or simply panic in case there is already an account for the generated key.

matthiasgeihs (approved):

PR #33, opened 2021-03-29
Add current State to HandleUpdate.

### LINKED ISSUE #15: Current state in ChannelUpdate
**Location**&emsp; `[client]`  

**Changes**&emsp; Add field `CurrentState` to `client.ChannelUpdate` to make the current state of a channel accessible.  
For version 1 updates it could either be the same as `State` or `null`.

**Context**&emsp; `channel.State()` can not be called in an ongoing `HandleUpdate` call.  
This makes it harder to use, since the user has to track the current state themself. Imagine:  
```go
func (c *Client) HandleUpdate(update client.ChannelUpdate, _ *client.UpdateResponder) {
    channel := c.getChannel(update.State.ID)
    current := channel.State()   // NOT POSSIBLE
}
```
It is not possible since the mutex of the machine that is associated with the channel is already locked.

**Acceptance criteria**&emsp; The state of a channel is accessible inside of an ongoing update in a threadsafe way.

ggwpez:
Since Channel.State() can not be called from an Update handler and
having the current state in the ChannelUpdate does not make sense,
i moved added it as argument to HandleUpate.

Closes #15

### DIFF
--- a/client/update.go
+++ b/client/update.go
@@ -1,4 +1,4 @@
-// Copyright 2019 - See NOTICE file for copyright holders.
+// Copyright 2021 - See NOTICE file for copyright holders.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -74,14 +74,16 @@ type (
 	// from other channel participants.
 	UpdateHandler interface {
 		// HandleUpdate is the user callback called by the channel controller on an
-		// incoming update request.
-		HandleUpdate(ChannelUpdate, *UpdateResponder)
+		// incoming update request. The first argument contains the current state
+		// of the channel before the update is applied. Clone it if you want to
+		// modify it.
+		HandleUpdate(*channel.State, ChannelUpdate, *UpdateResponder)
 	}
 
 	// UpdateHandlerFunc is an adapter type to allow the use of functions as
 	// update handlers. UpdateHandlerFunc(f) is an UpdateHandler that calls
 	// f when HandleUpdate is called.
-	UpdateHandlerFunc func(ChannelUpdate, *UpdateResponder)
+	UpdateHandlerFunc func(*channel.State, ChannelUpdate, *UpdateResponder)
 
 	// The UpdateResponder allows the user to react to the incoming channel update
 	// request. If the user wants to accept the update, Accept() should be called,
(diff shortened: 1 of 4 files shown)


manoranjith:
@ggwpez Why do you think it does not make sense to add the `CurrentState` to the `ChannelUpdate` struct ?

From the framework user point of view, I see that it is easy to propagate the `CurrentState` and `ProposedState` of the channel to the module that will have to decide whether to accept/reject the update.

client/update.go
```diff
@@ -276,7 +279,7 @@ func (c *Channel) handleUpdateReq(
 		return
 	}
 
-	uh.HandleUpdate(req.ChannelUpdate, responder)
+	uh.HandleUpdate(c.machine.State(), req.ChannelUpdate, responder)
```
matthiasgeihs, line 282: I think we should not provide a reference to the actual state, but only a reference to a clone: `c.machine.State().Clone()`, right?

matthiasgeihs (changes requested):

matthiasgeihs:
@ggwpez please also have a look at Mano's comment. What do you think?

ggwpez:
@manoranjith It does make sense to do it, but it would imply many changes in the framework code because.  
We internally use the struct and pass it around a lot. `ChannelUpdate` would then also not fulfill the `Encode/Decode` pattern anymore since the `CurrentState` should never be sent over the wire.

client/update.go
ggwpez: `Channel.State` also returns a pointer to the machine state, but i can clone it if you want.

manoranjith:
> @manoranjith It does make sense to do it, but it would imply many changes in the framework code because.
> We internally use the struct and pass it around a lot. `ChannelUpdate` would then also not fulfill the `Encode/Decode` pattern anymore since the `CurrentState` should never be sent over the wire.

Okay. 

If making it a part of the `ChannelUpdate` would be inconsistent with the current design of `Encoder/Decoder` pattern. I think, we could still comply with `Encoder/Decoder` pattern by excluding the current state when `Encoding/Decoding` and add it by reading from state machine after `Decoding`. But this would mean redundantly storing the current state and passing it around. 

Since the only purpose of this change is to make the current state available in `HandleUpdate` function, I think the current solutions of passing it as an additional parameter makes more sense.

matthiasgeihs (approved):

Issue #19, opened 2021-03-10
☂️ Reorg resistance of Ethereum backend


matthiasgeihs:
The Ethereum backend is currently build on the assumption that a transaction cannot be reverted. While this may be true for local testnets, it is not true for public testnets and the Ethereum mainnet, where block reorgs can happen.

We should make the Ethereum backend work with Ethereum nets where reorgs appear. An idea how to implement this is as follows: We introduce a parameter `ConfirmationBlockDepth`. Whenever go-perun accesses on-chain state, we only provide on-chain state if it is at least `ConfirmationBlockDepth` blocks old.

The following appear to be relevant places in the code where on-chain state is accessed:
- Contract validation: ValidateAdjudicator, ValidateAssetHolder
- Event subscription: Watch, Filter
- Contract deployment: WaitDeployed
- Transaction confirmation: ConfirmTransaction

Added by @ggwpez:  
The current road map for this issue looks like this:  
- [x] Reorg resistance in *go-perun*
  - [x] Simulated reorgs in *go-ethereum* https://github.com/ethereum/go-ethereum/pull/22624
    - [x] Tests
  - [x] Resistant Event Subs #40
    - [x] Use in *go-perun* #85
    - [x] Expose BoundContract #41 (done but not merged)
    - [x] Generic Event Subs #39 (done but not merged)
      - [x] Use in *go-perun* #84 
  - [x] Resistant TX confirming 
    - [x] Generic TX confirming #97 (partially done with `ConfirmTx` and `WaitDeployed`)
    - [x] Resistant contract deployment #99
    - [x] Fix finality depth #98
  - [x] Reorg Tests
    - [x] Reverting
    - [x] Reordering Events
    - [x] Reordering TX
    - [x] Header subscription

ggwpez:
Biggest point missing: How to test this.  
Currently we use the simulated block chain from `go-ethereum`. Im evaluating if it can do reorgs [here](https://github.com/hyperledger-labs/go-perun/compare/dev...perun-network:tmp/test-revert#diff-5008ce6fb9709c025c99e0bd5c3175fe257ad88a1114bf07b68f239e5f87a475R129).

ggwpez:
After adding a `Reorg(depth)` function to *go-ethereum*, i was able to get a legit reorg on the simulated backend.  
It removed the previous events with a `removed: true` event.  
It does seem, however, that the TX do not end up in the mempool again, and therefore needs to be resent by the user.   
[...]

PR #36, opened 2021-04-06
Generic events sub

ggwpez:
Add generic event subs in preparation for https://github.com/hyperledger-labs/go-perun/issues/19.

### DIFF
--- a/backend/ethereum/channel/funder.go
+++ b/backend/ethereum/channel/funder.go
@@ -27,7 +27,7 @@ import (
 	"github.com/pkg/errors"
 
 	"perun.network/go-perun/backend/ethereum/bindings/assetholder"
-	cherrors "perun.network/go-perun/backend/ethereum/channel/errors"
+	"perun.network/go-perun/backend/ethereum/subscription"
 	"perun.network/go-perun/backend/ethereum/wallet"
 	"perun.network/go-perun/channel"
 	"perun.network/go-perun/client"
@@ -184,7 +184,7 @@ func (f *Funder) sendFundingTx(ctx context.Context, request channel.FundingReq, 
 	// nolint: gocritic
 	if bal == nil || bal.Sign() <= 0 {
 		f.log.WithFields(log.Fields{"channel": request.Params.ID(), "idx": request.Idx}).Debug("Skipped zero funding.")
-	} else if alreadyFunded, err := checkFunded(ctx, bal, contract, fundingID); err != nil {
+	} else if alreadyFunded, err := f.checkFunded(ctx, bal, contract, fundingID); err != nil {
 		return nil, errors.WithMessage(err, "checking funded")
 	} else if alreadyFunded {
 		f.log.WithFields(log.Fields{"channel": request.Params.ID(), "idx": request.Idx}).Debug("Skipped second funding.")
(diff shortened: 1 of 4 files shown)


backend/ethereum/glue/doc.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package glue glues together go-perun and go-ethereum.
+// It contains generic functionality and can he used even without go-perun.
+package glue // import "perun.network/go-perun/backend/ethereum/glue"
```
matthiasgeihs, line 17: The package name "glue" is very vague. Use package name "subscription" instead?

backend/ethereum/glue/eventsub.go
```diff
@@ -0,0 +1,201 @@
...
+		closed   chan struct{}
+		contract *bind.BoundContract
+		eFact    EventFactory
+
+		watchLogs, filterLogs chan types.Log
+		watchSub, filterSub   event.Subscription
+	}
+
+	// EventType describes the type of an event.
+	EventType struct {
```
matthiasgeihs, line 39: It's confusing that the Data is part of the type.

Option 1: Merge EventType into Event.
Option 2: Move Data to Event.

It would be good to add a comment for each of the properties to explain what they stand for.

backend/ethereum/glue/eventsub.go
```diff
@@ -0,0 +1,201 @@
...
+			return nil
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			break read2
+		}
+	}
+
+	for _, log := range logs {
+		eventType := s.eFact()
```
matthiasgeihs, line 144: Here, you should be generating an event, not just an event type.

backend/ethereum/glue/eventsub.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	// Event is a generic event as received from go-ethereum.
+	Event struct {
+		EventType
+		Log types.Log
+	}
+
+	// EventFactory is used to create `EventType`s.
+	EventFactory func() EventType
```
matthiasgeihs, line 52: The event factory should return an Event, not just an EventType.

backend/ethereum/glue/eventsub.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+	}
+	return nil
+}
+
+func (s *EventSub) readFuture(ctx context.Context, sink chan<- *Event) error {
+	for {
+		select {
+		case log := <-s.watchLogs:
+			eventType := s.eFact()
```
matthiasgeihs, line 168: Same as above, should return Event, not EventType.

backend/ethereum/glue/eventsub_test.go
```diff
@@ -0,0 +1,118 @@
...
+			require.NoError(t, err)
+			tx, err := token.IncreaseAllowance(opts, account.Address, big.NewInt(1))
+			require.NoError(t, err)
+			// Wait for the TX to be mined.
+			_, err = cb.ConfirmTransaction(ctx, tx, *account)
+			require.NoError(t, err)
+		}
+	})
+	// Random wait to get some events before and some after the event sub is setup.
+	time.Sleep(time.Duration(rng.Int31n(500)+100) * time.Millisecond)
```
matthiasgeihs, line 85: Do we know how many events will be emitted on average before the subscription is setup? I think it would be better to deterministically ensure that at least a small amount of events have been emitted before the setup.

matthiasgeihs (changes requested):
Please push your changes as new commits as long as the review is in progress. Do not force push until the code is stable.

ggwpez:
> You may consider integrating into this PR already some of the places where the subscription will be used. Otherwise it could happen that we only realize necessary changes later.

Ok i will try it out at some spots where we use it.

backend/ethereum/bindings/peruntoken/PerunToken.go
```diff
@@ -380,6 +380,10 @@ type ERC20Caller struct {
 	contract *bind.BoundContract // Generic contract wrapper for the low level calls
 }
 
+func (e *ERC20Caller) Contract() *bind.BoundContract {
+	return e.contract
```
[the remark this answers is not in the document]
ggwpez: Did not think about this before! good idea

backend/ethereum/glue/eventsub.go
ggwpez: Ok i think will merge them. Should be more understandable that way.

ggwpez:
While using the `EventSub` in the Funder, it seems that the `start, end *uint64` pointers that specify the range to query are not easy to use.  
I will try to change it just to `pastBlocks` which is then a relative offset into the past.

ggwpez:
Since the `EventSub` is using geth functions, we need to check the `ChainNotReachableError` for every returned error.  
The function `isChainNotReachableError` is in `[backend/eth/channel]` (and private), which prevents me from importing it, since it would create a cyclic dependency.  

So we either make an addictional `[backend/eth/channel/error]` package or merge the `glue` package into  `[backend/eth/channel]`.  
Any other ideas @matthiasgeihs ?

backend/ethereum/glue/doc.go
ggwpez: Renamed it to `subscription` for now. I also wanted to put the reorg-resistant TX stuff in here, then the name `subscription` would not be enough.

backend/ethereum/bindings/peruntoken/PerunToken.go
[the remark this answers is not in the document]
ggwpez: I added `abi.go` now.

backend/ethereum/glue/eventsub_test.go
ggwpez: I added a sync channel which allows the reader to wait until `n/4` TX were sent.

matthiasgeihs:
> Since the `EventSub` is using geth functions, we need to check the `ChainNotReachableError` for every returned error.
> The function `isChainNotReachableError` is in `[backend/eth/channel]` (and private), which prevents me from importing it, since it would create a cyclic dependency.
> 
> So we either make an addictional `[backend/eth/channel/error]` package or merge the `glue` package into `[backend/eth/channel]`.
> Any other ideas @matthiasgeihs ?

I don't understand. Why do we need to check for `ChainNotReachableError` because of using geth functions? Isn't the error type specific to go-perun?

Anyways, adding a package seems cleaner to me. Or could we maybe check for the error without using a dedicated function?

backend/ethereum/channel/funder.go
```diff
@@ -184,7 +184,7 @@ func (f *Funder) sendFundingTx(ctx context.Context, request channel.FundingReq,
 	// nolint: gocritic
 	if bal == nil || bal.Sign() <= 0 {
 		f.log.WithFields(log.Fields{"channel": request.Params.ID(), "idx": request.Idx}).Debug("Skipped zero funding.")
-	} else if alreadyFunded, err := checkFunded(ctx, bal, contract, fundingID); err != nil {
+	} else if alreadyFunded, err := f.checkFunded(ctx, bal, contract, fundingID); err != nil {
```
matthiasgeihs, line 187: Not sure why we check the funding here before we actually do it, but this should probably be not the subject of the current PR.

backend/ethereum/channel/funder.go
```diff
@@ -210,76 +210,61 @@ func (f *Funder) deposit(ctx context.Context, bal *big.Int, asset Asset, funding
...
-
-	return iter, nil
+	event := func() *subscription.Event {
+		return &subscription.Event{
+			Name:   "Deposited",
+			Data:   new(assetholder.AssetHolderDeposited),
+			Filter: [][]interface{}{filter},
+		}
+	}
+	sub, err := subscription.NewEventSub(ctx, f, contract, event, startBlockOffset)
```
matthiasgeihs, line 248: There is a difference in functionality here compared to before. Before, we were starting from block 1. Here we are starting from current block - offset. I think it's OK because we also do it in other places but in general this is an open issue how to choose the startBlockOffset correctly.

matthiasgeihs (approved):
Only general comments. Looks good.

backend/ethereum/channel/funder.go
ggwpez: I will open an issue for brainstorming it.

backend/ethereum/channel/funder.go
ggwpez: The framework could be restarted after sending the `deposit` transactions, in which case we do not want to send them twice.  
It is still a race condition in that case, but should mitigate most of it. This is what the docu means with `It is idempotent.`.  
@matthiasgeihs

matthiasgeihs (approved):

PR #37, opened 2021-04-06
WIP: Test geth sim reorgs

ggwpez:
Test the assumptions on which we base the reorg resistance using simulated reorgs from geth.


ggwpez:
Will be part of #106, closing.

PR #53, opened 2021-04-20
[backend/eth/channel] Retry WaitDeployed in deployment.

### LINKED ISSUE #52: Contract deployment on Arbitrum
**Location**: `[backend/eth/channel] deploy.go`  

**Problem**: While using the *Kovan4* endpoint of *Arbitrum*, i sometimes got an error when deploying our contracts:  
`setting up contracts: contract setup failed: deploying eth assetholder: deploying ETHAssetHolder: no contract code after deployment`.  
*go-ethereum* queries the code of a deployed contract immediately after deploying it.  
It seems that either by design or because *go-ethereum* just waits one block after deploying, it sometimes returns no code.

**Solution**: *go-etherum* returns an `ErrNoCodeAfterDeploy` error in this case.  
Checking for it in a retry loop with 3 tries and a 2 seconds delay should make up for it.  

**Testing**: I would test it again with the demo client.

ggwpez:
While trying out Arbitrum it was necessary to add this delay and
retry since it would otherwise report 'No Code' after deployment.

Closes #52

### DIFF
--- a/backend/ethereum/channel/deploy.go
+++ b/backend/ethereum/channel/deploy.go
@@ -16,8 +16,8 @@ package channel
 
 import (
 	"context"
-	"fmt"
 	"math/big"
+	"time"
 
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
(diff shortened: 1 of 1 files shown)


ggwpez:
Out of scope.

PR #55, opened 2021-04-27
:bug: [client] Fix NewRandomLedgerChannelProposal

matthiasgeihs:
An inconsistent LedgerChannelProposal may have been generated because
the allocation was generated independent of the number of participants.


ggwpez:
This is indeed a bug, but it was currently prevented by `NewRandomBaseChannelProposal` internally calling `opt.NumParts(rng)` on the passed option which would then fix the number of participants.  
The second call here to `opt.NumParts(rng)` is therefore equivalent to `base.NumParts()`.

ggwpez (approved):

PR #59, opened 2021-05-03
:bug: [channel/machine, client] Sub-channel withdrawal without regist…

### LINKED ISSUE #58: Sub-channel settlement requires channel registration on ajdudicator
In the optimistic case we can settle a subchannel into its parent channel without having to go on-chain. However, our state machine logic currently does not allow to settle and withdraw a channel from the final state (no valid transition from phase final to phase withdrawing). We should change the state machine behavior so that it allows settling subchannels without channel registration on-chain.

matthiasgeihs:
…ering

The current implementation required a sub-channel to be registered on-chain before it could be withdrawn.
However, in the optimistic case, a sub-channel should be withdrawable into its parent channel without
blockchain interaction. We achieve this by modifying the state machine to allow for a transition from
phase final to phase withdrawing.

Closes #58

### DIFF
--- a/channel/machine.go
+++ b/channel/machine.go
@@ -400,10 +400,10 @@ func (m *machine) SetProgressed(e *ProgressedEvent) error {
 
 // SetWithdrawing sets the state machine to the Withdrawing phase. The current
 // state was registered on-chain and funds withdrawal is in progress.
-// This phase can only be reached from phase Acting, Registered, Progressed, or
+// This phase can only be reached from phase Final, Registered, Progressed, or
 // Withdrawing.
 func (m *machine) SetWithdrawing() error {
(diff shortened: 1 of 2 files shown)


ggwpez:
I dont quite get the change.  
You say that you change it for sub-channels, but the change in the `machine` also affects ledger channels, or not?

matthiasgeihs:
It affects all channels. The situation after the change is not perfect. But it is better than before. In the long run, we could reconsider our machine logic as a whole.

ggwpez:
> It affects all channels. The situation after the change is not perfect. But it is better than before. In the long run, we could reconsider our machine logic as a whole.

Ok.

After your change it is not possible anymore to `Settle` a channel before having called `Register`, which is reasonable IMHO.

ggwpez (approved):

ggwpez (approved):

PR #60, opened 2021-05-03
:lock: [channel] Check ID after decoding parameters

### LINKED ISSUE #56: Params.Decode may create invalid parameters
Parameters are not checked for consistency when created by deserialization. In particular, it is not checked whether the ID is correct. However, this may be crucial for the security of our program. Hence, we must check that `params.id == CalcID(params)` after deserialization.

matthiasgeihs:
When deserializing parameters, it must be checked whether the ID is
correct. Otherwise we might end up with an inconsistent Params object.

Closes #56

### DIFF
--- a/channel/params_test.go
+++ b/channel/params_test.go
@@ -34,9 +34,11 @@ func TestParams_Serializer(t *testing.T) {
 	rng := pkgtest.Prng(t)
 	params := make([]io.Serializer, 10)
 	for i := range params {
-		p := test.NewRandomParams(rng)
+		var p *channel.Params
 		if i&1 == 0 {
-			p.App = channel.NoApp()
+			p = test.NewRandomParams(rng, test.WithoutApp())
+		} else {
+			p = test.NewRandomParams(rng)
 		}
 		params[i] = p
 	}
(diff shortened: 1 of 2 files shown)


ggwpez:
There is also the possibility to not send the `ID` and just calculate it in the *decode* step, right?  
Why did you opt for this instead? @matthiasgeihs

channel/params_test.go
```diff
@@ -34,9 +34,11 @@ func TestParams_Serializer(t *testing.T) {
 	rng := pkgtest.Prng(t)
 	params := make([]io.Serializer, 10)
 	for i := range params {
-		p := test.NewRandomParams(rng)
+		var p *channel.Params
 		if i&1 == 0 {
-			p.App = channel.NoApp()
+			p = test.NewRandomParams(rng, test.WithApp(channel.NoApp()))
```
ggwpez, line 39: We have `WithoutApp` for this case which also correctly sets `WithAppData`.

matthiasgeihs:
> There is also the possibility to not send the `ID` and just calculate it in the _decode_ step, right?
> Why did you opt for this instead? @matthiasgeihs

I wanted to do it with minimal changes. Anyways, edited the code now based on your suggestion.

channel/params_test.go
matthiasgeihs: done

ggwpez (approved):

PR #61, opened 2021-05-03
:lock: [client] Verify state signatures for subchannel funding and se…

### LINKED ISSUE #57: Client does not verify signatures for subchannel funding and withdrawal updates
In update.go, `machine.CheckUpdate` is only called after handling subchannel funding and withdrawal updates. Hence, the check is not done in these cases. This may lead the client to accepting an update where the signatures are invalid. `machine.CheckUpdate` must be called before any response.

matthiasgeihs:
…ttlement updates

We are intercepting updates for subchannel funding and settlement. However,
we did not check the signatures for these updates. This is fixed now.

Closes #57

### DIFF
--- a/client/update.go
+++ b/client/update.go
@@ -267,6 +267,12 @@ func (c *Channel) handleUpdateReq(
 	c.machMtx.Lock() // Lock machine while update is in progress.
 	defer c.machMtx.Unlock()
 
+	if err := c.machine.CheckUpdate(req.State, req.ActorIdx, req.Sig, pidx); err != nil {
+		// TODO: how to handle invalid updates? Just drop and ignore them?
+		c.logPeer(pidx).Warnf("invalid update received: %v", err)
+		return
+	}
+
 	responder := &UpdateResponder{channel: c, pidx: pidx, req: req}
 
 	if ui, ok := c.subChannelFundings.Filter(req.ChannelUpdate); ok {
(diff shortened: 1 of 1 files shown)


client/update.go
```diff
@@ -267,6 +267,12 @@ func (c *Channel) handleUpdateReq(
 	c.machMtx.Lock() // Lock machine while update is in progress.
 	defer c.machMtx.Unlock()
 
+	if err := c.machine.CheckUpdate(req.State, req.ActorIdx, req.Sig, pidx); err != nil {
```
ggwpez, line 270: It seems alright to move this up here, but we now do not check anymore for `pidx == req.ActorIDx`.  
To keep the protocol simple we should also put that check here.  
Otherwise a sub-channel proposer could put the proposee as Actor for the update.  

The check is currently done in `validTwoPartyUpdate`.

ggwpez (changes requested):

client/update.go
matthiasgeihs: Can you give an example what would be problematic with this? I suggest you make a separate issue for that.

ggwpez: Alice can now send a `subChannelFundings` or `subChannelWithdrawals` to Bob and put Bobs index as *actorIDx*.  
For Bob it now looks like he received an update from himself.  

I think it is not a security issue, but still unclean.

matthiasgeihs: Yes, it's not clean, I agree. On the other hand, it is a separate issue, so you may want to create one. I had a look into it and it may require some careful thinking.

The issue addressed here is more critical and I would like to get the change through without too much delay.

ggwpez (approved):

PR #70, opened 2021-05-07
Check for nil parts in ValidateParams.

### LINKED ISSUE #69: [channel] ValidateParameters check parts
**Problem:** `[channel] params.go`  
Function `ValidateParameters` does not check for `nil` values in the participants slice since it is only passed `len(parts)` instead of `parts`.  
This can subsequently cause a panic in `CalcID` which assumes that all values are valid.

**Solution:** Pass the participants as `parts []wallet.Address` to `ValidateParameters` and check that none of the values is `nil`.  
As additional feature it could check that none of the participants is the zero address.

ggwpez:
Closes #69

### DIFF
--- a/channel/params.go
+++ b/channel/params.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	stdio "io"
 	"math/big"
+	"reflect"
 
 	"github.com/pkg/errors"
 
@@ -79,7 +80,7 @@ func (p *Params) ID() ID {
 // is also calculated here and persisted because it probably is an expensive
 // hash operation.
 func NewParams(challengeDuration uint64, parts []wallet.Address, app App, nonce Nonce) (*Params, error) {
-	if err := ValidateParameters(challengeDuration, len(parts), app, nonce); err != nil {
+	if err := ValidateParameters(challengeDuration, parts, app, nonce); err != nil {
 		return nil, errors.WithMessage(err, "invalid parameter for NewParams")
 	}
 	return NewParamsUnsafe(challengeDuration, parts, app, nonce), nil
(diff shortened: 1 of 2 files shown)


matthiasgeihs (approved):

channel/params.go
```diff
@@ -118,6 +119,11 @@ func ValidateParameters(challengeDuration uint64, numParts int, app App, nonce N
 	if len(nonce.Bytes()) > MaxNonceLen {
 		return errors.Errorf("nonce too long (%d > %d)", len(nonce.Bytes()), MaxNonceLen)
 	}
+	for i, part := range parts {
+		if part == nil {
```
matthiasgeihs, line 123: There are some particularities about checking interfaces against nil. Can you justify that doing that is ok here?

https://stackoverflow.com/questions/13476349/check-for-nil-and-nil-interface-in-go

ggwpez: I did not think about this, looks like `(wallet.Address)nil` would not be detected as `nil` correctly.  
Will look into it :smile:

matthiasgeihs: Maybe the correct way of dealing with this would be to ensure that we never construct Params that are invalid?

ggwpez: We do have `NewParamsUnsafe` which possibly does that.  
Otherwise when decoding `Params`, we do not check for validity of the decoded values.  

I will add a test with reflection to check for nil, since golang does not seem to offer any other possibility of checking that.

matthiasgeihs: It is advised to avoid using package reflect when possible and I think this case is not important enough to use it here.

In fact, it is currently unclear to me why we need to check for this at all? Isn't it the problem of the framework user if it sets a participant address to nil? Why would we need to protect against this?

matthiasgeihs: @ggwpez hm? can you explain where exactly you were running in problems if you don't check for nil participants? Otherwise I would close the issue.

ggwpez: Well, the problem that i tried to fix is described in #69. But it seems not possible without the use reflection. @matthiasgeihs

matthiasgeihs: #69 only describes a potential problem, but I have trouble understanding if this is an actual problem or it is rather artificial. This is why I wrote above:

> In fact, it is currently unclear to me why we need to check for this at all? Isn't it the problem of the framework user if it sets a participant address to nil? Why would we need to protect against this?

ggwpez: It is something that i noticed when changing code for the hackathon.  
I dont know if there is currently a problem with it, i just thought that it is weird if `ValidateParams` does not actually validate the params.

matthiasgeihs: OK, so it seems that there is not an actual problem with it for now?

I am not super convinced that we need it, especially since it would be the first usage of package reflect.

ggwpez: Well ok then lets wait for it to pops up somewhere again.

ggwpez:
Not important urgent right now, closing.

PR #71, opened 2021-05-07
:sparkles: [channel] Improve Allocation handling

### LINKED ISSUE #63: Easier method for creating and handling an allocation
At the moment, creating a channel allocation for channel opening is not very convenient and can be confusing. Even for just a single asset channel it looks like the following.
```
channel.Allocation{
	Assets:   []channel.Asset{asset},
	Balances: [][]*big.Int{etherToWei(myBalEth, peerBalEth)},
}
```

We may want to change the Allocation struct and provide helper methods for that.

The allocation struct could hold a `map[Asset]AccountBalances`, where AccountBalances is a map from accounts to integers.

matthiasgeihs:
One step towards making it more convenient to work with `Allocation`.

Closes #63

### DIFF
--- a/channel/allocation.go
+++ b/channel/allocation.go
@@ -21,6 +21,7 @@ import (
 
 	perunio "perun.network/go-perun/pkg/io"
 	perunbig "perun.network/go-perun/pkg/math/big"
+	"perun.network/go-perun/wire"
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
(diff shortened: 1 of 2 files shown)


ggwpez (changes requested):
Sorry that i did not review this earlier @matthiasgeihs, seems like i missed in my github page.
  
PS: It is still in draft.

matthiasgeihs:
Closed due to inactivity

PR #74, opened 2021-05-13
Simplify funder interface

### LINKED ISSUE #72: Simplify Funder interface
It is currently rather complicated to setup a funder.

The constructor `NewFunder` takes as input `accounts map[Asset]accounts.Account` and `depositors map[Asset]Depositor`, both mapping from the same list of assets to a set of objects. For ease of use, these two maps could be combined into one map by integrating the account into the depositor.

Furthermore, there are no constructors for these maps and not much documentation on `NewFunder`, so it is hard to know how exactly the maps should be created. For ease of use, we should extend the documentation of `NewFunder` and we should make it easier to create a depositor map, e.g., by adding a special type for it with respective helper methods.

matthiasgeihs:
Closes #72

### DIFF
--- a/backend/ethereum/channel/erc20_depositor.go
+++ b/backend/ethereum/channel/erc20_depositor.go
@@ -36,6 +36,11 @@ type ERC20Depositor struct {
 // An `IncreaseAllowance` uses ~45kGas and a `Deposit` call ~84kGas on average.
 const ERC20DepositorTXGasLimit = 100000
 
+// NewERC20Depositor creates a new ERC20Depositor.
+func NewERC20Depositor(token common.Address) *ERC20Depositor {
+	return &ERC20Depositor{Token: token}
+}
+
 // Deposit deposits ERC20 tokens into the ERC20 AssetHolder specified at the
 // requests's asset address.
 func (d *ERC20Depositor) Deposit(ctx context.Context, req DepositReq) (types.Transactions, error) {
--- a/backend/ethereum/channel/eth_depositor.go
+++ b/backend/ethereum/channel/eth_depositor.go
@@ -33,6 +33,11 @@ type ETHDepositor struct{}
 // A `Deposit` call uses ~47kGas on average.
 const ETHDepositorGasLimit = 50000
 
+// NewETHDepositor creates a new ETHDepositor.
+func NewETHDepositor() *ETHDepositor {
+	return &ETHDepositor{}
+}
+
 // Deposit deposits ether into the ETH AssetHolder specified at the requests's
 // asset address.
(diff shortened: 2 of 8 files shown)


backend/ethereum/channel/depositor.go
```diff
@@ -27,6 +27,9 @@ type (
 	// Depositor is used by the `Funder` to deposit funds for ledger channel
 	// funding. Depositor are reusable such that one Depositor per asset is enough.
 	Depositor interface {
+		// Account returns the depositor account.
+		Account() accounts.Account
```
ggwpez, line 31: If you want to store the account in the `Depositor`, you could also remove it from the `DepositReq` and remove this function.

ggwpez:
I dont really like the concept that the `Depositor` is now additionally used as an account storage.  
You could also change the `DepositorMap` to a map which holds *both* a `Depositor` and an `Account`. Then you would not need to change much in the logic of the `Depositor`s.

backend/ethereum/channel/depositor.go
```diff
@@ -27,6 +27,9 @@ type (
 	// Depositor is used by the `Funder` to deposit funds for ledger channel
 	// funding. Depositor are reusable such that one Depositor per asset is enough.
```
ggwpez, line 28: The comment was designed for stateless `Depositor`s, so this is wrong now.

backend/ethereum/channel/eth_depositor.go
```diff
@@ -26,13 +27,25 @@ import (
 
 // ETHDepositor deposits funds into the `AssetHolderETH` contract.
 // It has no state and is therefore completely reusable.
```
ggwpez, line 29: Comment out-dated.

backend/ethereum/channel/erc20_depositor.go
```diff
@@ -29,13 +30,27 @@ import (
 // It is bound to a token but can be reused to deposit multiple times.
```
ggwpez, line 30: It is now also bound to an `Account`.

ggwpez (changes requested):

matthiasgeihs:
> I dont really like the concept that the `Depositor` is now additionally used as an account storage.
> You could also change the `DepositorMap` to a map which holds _both_ a `Depositor` and an `Account`. Then you would not need to change much in the logic of the `Depositor`s.

Why don't you like it?

At first I was considering to do as you suggest (i.e., keep Depositor and Account separate and use a map from Asset to struct{Depositor, Account}). However, after thinking about it, I didn't see a reason why they should be separated. Logically, it makes sense to me that when you create a depositor, you also specify the account for that depositor. The construction of a funder is also more convenient that way, as you need to construct one object less than before.

PR #80, opened 2021-05-16
:boom: [channel] Update NewFundingTimeoutError to return a value

### LINKED ISSUE #73: [client] NewFundingTimeoutErrror can return error as value instead of pointer.
Currently, `NewFundingTimeoutErrror`, the constructor for error type `FundingTimeoutError` [returns a pointer](https://github.com/hyperledger-labs/go-perun/blob/bf29b2bf504f57261e04601e31e3a3cf9cafee43/channel/funder.go#L76) instead of value. This makes the usage of the error type tricky.

For example, the [`client.Proposal`](https://github.com/hyperledger-labs/go-perun/blob/bf29b2bf504f57261e04601e31e3a3cf9cafee43/client/proposal.go#L145) returns `PeerRejectedError`, `RequestTimedoutError`, `FundingTimedoutError` and few more.

Since each of the other errors are generated as values by their respective constructors, then can be retrieved as below:
```
peerRejectedError := pclient.PeerRejectedError{}
ok := errors.As(err, &peerRejectedError)
```

But for `FundingTimeoutError`, a double pointer should be used
```
fundingTimeoutError := &pchannel.FundingTimeoutError{}
ok := errors.As(err, &fundingTimeoutError)
```

Is there a specific reason for returning a pointer in this case ?

If not, I think constructor of `FundingTimeoutError` can be updated to return a value, so that it is more idiomatic and does not requires an exception when handling the error.

manoranjith:
- Previously, the function returned pointer to an error. Update it to
  return a value.

- It is idiomatic to return error types as values. The change also
  ensures this func is consistent with constructors for other error
  types.

Closes #73.

### DIFF
--- a/backend/ethereum/channel/funder_test.go
+++ b/backend/ethereum/channel/funder_test.go
@@ -251,7 +251,7 @@ func testFundingTimeout(t *testing.T, faultyPeer, n int) {
 			err := funder.Fund(ctx, *req)
 			require.Error(t, err)
 			require.True(rt, channel.IsFundingTimeoutError(err), "funder should return FundingTimeoutError")
-			pErr := errors.Cause(err).(*channel.FundingTimeoutError) // unwrap error
+			pErr := errors.Cause(err).(channel.FundingTimeoutError) // unwrap error
 			// Check that `faultyPeer` is reported as faulty.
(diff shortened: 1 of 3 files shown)


ggwpez (approved):
Thanks :pray:

PR #89, opened 2021-05-31
Use EventSub instead of Filter/Watch.

### LINKED ISSUE #84: Use generic EventSubs in eth/channel
**Location**: `[backend/eth/channel]`  

**Motivation**: After #39 is done, the generic `EventSub` can be used in all places where `Watch…` or `Filter…` is used currently to monitor events. This also helps to finalize the design of #39, which is currently in review.    

**Testing**: The tests should not need any modification.

ggwpez:
Closes #84

### DIFF
--- a/backend/ethereum/channel/conclude.go
+++ b/backend/ethereum/channel/conclude.go
@@ -21,8 +21,10 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/pkg/errors"
 
+	"perun.network/go-perun/backend/ethereum/bindings"
 	"perun.network/go-perun/backend/ethereum/bindings/adjudicator"
 	cherrors "perun.network/go-perun/backend/ethereum/channel/errors"
+	"perun.network/go-perun/backend/ethereum/subscription"
 	"perun.network/go-perun/channel"
 )
 
@@ -31,35 +33,36 @@ const secondaryWaitBlocks = 2
 // ensureConcluded ensures that conclude or concludeFinal (for non-final and
 // final states, resp.) is called on the adjudicator.
 // - a subscription on Concluded events is established
-// - it searches for a past concluded event
+// - it searches for a past concluded event by calling `isConcluded`
 //   - if found, channel is already concluded and success is returned
 //   - if none found, conclude/concludeFinal is called on the adjudicator
 // - it waits for a Concluded event from the blockchain.
 func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.AdjudicatorReq, subStates channel.StateMap) error {
-	// Listen for Concluded event.
-	watchOpts, err := a.NewWatchOpts(ctx)
+	sub, err := subscription.NewEventSub(ctx, a.ContractBackend, a.bound, updateEventType(req.Params.ID()), startBlockOffset)
 	if err != nil {
-		return errors.WithMessage(err, "creating watchOpts")
+		return errors.WithMessage(err, "subscribing")
 	}
-	events := make(chan *adjudicator.AdjudicatorChannelUpdate)
-	sub, err := a.contract.WatchChannelUpdate(watchOpts, events, [][32]byte{req.Params.ID()})
-	if err != nil {
-		err = cherrors.CheckIsChainNotReachableError(err)
-		return errors.WithMessage(err, "creating subscription failed")
-	}
-	defer sub.Unsubscribe()
-
-	if found, err := a.filterConcluded(ctx, req.Params.ID()); err != nil {
-		return errors.WithMessage(err, "filtering old Concluded events")
-	} else if found {
+	defer sub.Close()
+	// Check whether it is already concluded.
+	if concluded, err := a.isConcluded(ctx, sub); err != nil {
+		return errors.WithMessage(err, "isConcluded")
+	} else if concluded {
 		return nil
 	}
+
+	events := make(chan *subscription.Event, 10)
+	subErr := make(chan error, 1)
+	waitCtx, cancel := context.WithCancel(ctx)
+	go func() {
+		subErr <- sub.Read(ctx, events)
+		cancel()
+	}()
 
 	// In final Register calls, as the non-initiator, we optimistically wait for
 	// the other party to send the transaction first for secondaryWaitBlocks many
 	// blocks.
 	if req.Tx.IsFinal && req.Secondary {
-		isConcluded, err := waitConcludedForNBlocks(ctx, a, sub, events, secondaryWaitBlocks)
+		isConcluded, err := waitConcludedForNBlocks(waitCtx, a, events, secondaryWaitBlocks)
 		if err != nil {
 			return err
 		} else if isConcluded {
@@ -79,14 +82,51 @@ func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.Adjudicat
 		return err
 	}
 
-	select {
-	case <-events:
-		return nil
-	case <-ctx.Done():
-		return errors.Wrap(ctx.Err(), "context cancelled")
-	case err = <-sub.Err():
-		err = cherrors.CheckIsChainNotReachableError(err)
-		return errors.Wrap(err, "subscription error")
+	for {
+		select {
+		case _e := <-events:
+			e := _e.Data.(*adjudicator.AdjudicatorChannelUpdate)
+			if e.Phase == phaseConcluded {
+				return nil
+			}
+		case <-ctx.Done():
+			return errors.Wrap(ctx.Err(), "context cancelled")
+		case err = <-subErr:
+			if err != nil {
+				return errors.WithMessage(err, "subscription error")
+			}
+			return errors.New("subscription closed")
+		}
+	}
+}
+
+// isConcluded returns whether a channel is already concluded.
+func (a *Adjudicator) isConcluded(ctx context.Context, sub *subscription.EventSub) (bool, error) {
+	events := make(chan *subscription.Event, 10)
+	subErr := make(chan error, 1)
+	// Write the events into events.
+	go func() {
+		defer close(events)
+		subErr <- sub.ReadPast(ctx, events)
+	}()
+	// Read all events and check for concluded.
+	for _e := range events {
+		e := _e.Data.(*adjudicator.AdjudicatorChannelUpdate)
+		if e.Phase == phaseConcluded {
+			return true, nil
+		}
+	}
+	return false, errors.WithMessage(<-subErr, "reading past events")
+}
+
+func updateEventType(channelID [32]byte) subscription.EventFactory {
+	return func() *subscription.Event {
+		return &subscription.Event{
+			Name: bindings.Events.AdjChannelUpdate,
+			Data: new(adjudicator.AdjudicatorChannelUpdate),
+			// In the best case we could already filter for 'Concluded' phase only here.
+			Filter: [][]interface{}{{channelID}},
+		}
 	}
 }
 
@@ -98,11 +138,10 @@ func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.Adjudicat
 // the Concluded event subscription instance.
 func waitConcludedForNBlocks(ctx context.Context,
 	cr ethereum.ChainReader,
-	sub ethereum.Subscription,
-	concluded chan *adjudicator.AdjudicatorChannelUpdate,
+	concluded chan *subscription.Event,
 	numBlocks int,
 ) (bool, error) {
-	h := make(chan *types.Header)
+	h := make(chan *types.Header, 10)
 	hsub, err := cr.SubscribeNewHead(ctx, h)
 	if err != nil {
 		err = cherrors.CheckIsChainNotReachableError(err)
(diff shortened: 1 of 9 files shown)


backend/ethereum/channel/conclude.go
```diff
@@ -98,11 +128,10 @@ func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.Adjudicat
 // the Concluded event subscription instance.
 func waitConcludedForNBlocks(ctx context.Context,
 	cr ethereum.ChainReader,
-	sub ethereum.Subscription,
-	concluded chan *adjudicator.AdjudicatorChannelUpdate,
+	concluded chan *subscription.Event,
 	numBlocks int,
 ) (bool, error) {
-	h := make(chan *types.Header)
+	h := make(chan *types.Header, 10)
```
matthiasgeihs, line 134: Why with buffer now?

backend/ethereum/channel/subscription.go
```diff
@@ -17,115 +17,74 @@ package channel
...
 
 // Subscribe returns a new AdjudicatorSubscription to adjudicator events.
 func (a *Adjudicator) Subscribe(ctx context.Context, params *channel.Params) (channel.AdjudicatorSubscription, error) {
-	events := make(chan *adjudicator.AdjudicatorChannelUpdate)
-	sub, iter, err := a.filterWatch(ctx, events, params)
+	subErr := make(chan error, 1)
+	events := make(chan *subscription.Event, 10)
+	eFact := func() *subscription.Event {
+		return &subscription.Event{
+			Name:   "ChannelUpdate",
```
matthiasgeihs, line 39: Try to assert correct event name programmatically, see also comment above.

backend/ethereum/channel/conclude.go
```diff
@@ -69,24 +79,44 @@ func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.Adjudicat
...
+	}
+}
+
+func updateEventType(channelID [32]byte) subscription.EventFactory {
+	return func() *subscription.Event {
+		return &subscription.Event{
+			Name: "ChannelUpdate",
+			Data: new(adjudicator.AdjudicatorChannelUpdate),
+			// In the best case we could already filter for 'Concluded' phase only here.
+			Filter: [][]interface{}{{channelID}},
```
matthiasgeihs, line 118: Can we filter for `phase == Concluded`?

backend/ethereum/channel/subscription.go
```diff
@@ -17,115 +17,74 @@ package channel
...
+	closed sync.Once
 }
 
-// HasPast indicates whether there was a past event when the subscription was created.
-func (r *RegisteredSub) HasPast() bool {
-	return r.past
-}
-
-func (r *RegisteredSub) updateNext(ctx context.Context, events chan *adjudicator.AdjudicatorChannelUpdate, a *Adjudicator) {
+// nolint: gocognit
```
matthiasgeihs, line 74: Don't ignore warnings if not necessary. Try to simplify your function. Should be possible, no?

backend/ethereum/channel/subscription.go
```diff
@@ -17,115 +17,74 @@ package channel
...
+func (r *RegisteredSub) updateNext(ctx context.Context, events chan *subscription.Event, a *Adjudicator) {
```
matthiasgeihs, line 75: Why do we need such a complicated logic again? Couldn't we just use our subscription?

backend/ethereum/channel/withdraw.go
```diff
@@ -55,25 +56,24 @@ func (a *Adjudicator) ensureWithdrawn(ctx context.Context, req channel.Adjudicat
...
 			}
-			defer sub.Unsubscribe()
+			defer sub.Close()
 
-			// Filter past events
-			if found, err := a.filterWithdrawn(ctx, fundingID, contract); err != nil {
-				return errors.WithMessage(err, "filtering old Withdrawn events")
-			} else if found {
+			// Check for past event
+			if err := sub.ReadPast(ctx, events); err != nil {
```
matthiasgeihs, line 70: Why do you use ReadPast? Why not just Read?

backend/ethereum/channel/withdraw.go
```diff
@@ -82,20 +82,36 @@ func (a *Adjudicator) ensureWithdrawn(ctx context.Context, req channel.Adjudicat
...
 			}
 		})
 	}
 	return g.Wait()
 }
 
+func withdrawnEventType(fundingID [32]byte) subscription.EventFactory {
+	return func() *subscription.Event {
+		return &subscription.Event{
+			Name:   "Withdrawn",
```
matthiasgeihs, line 108: Assert or read event name programmatically.

backend/ethereum/subscription/eventsub.go
```diff
@@ -109,6 +109,8 @@ func calcStartBlock(ctx context.Context, chain ethereum.ChainReader, pastBlocks
 // Read reads all past and future events into `sink`.
 // Can be aborted by cancelling `ctx` or `Close()`.
 // It is possible that the same event is read more than once.
+// After casting the generic event to a specific type, the `Raw` log field
+// will be nil. Use the `Log` field of the generic event instead.
```
matthiasgeihs, line 113: Why does the Raw log field become nil? Is this intentional?

matthiasgeihs (commented):
The integration is a bit more complex than I hoped.

Couldn't we just use sub.Read without using sub.ReadPast in many cases?

Hardcoding the event name is not good practice.

backend/ethereum/channel/conclude.go
```diff
@@ -69,24 +79,44 @@ func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.Adjudicat
...
+			}
+			return errors.New("subscription closed")
+		}
+	}
+}
+
+func updateEventType(channelID [32]byte) subscription.EventFactory {
+	return func() *subscription.Event {
+		return &subscription.Event{
+			Name: "ChannelUpdate",
```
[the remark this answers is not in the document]
ggwpez: I can also add the events in the `abi.go` file to fix them.

backend/ethereum/channel/conclude.go
```diff
@@ -35,31 +36,40 @@ const secondaryWaitBlocks = 2
...
+		return errors.WithMessage(err, "subscribing")
 	}
-	events := make(chan *adjudicator.AdjudicatorChannelUpdate)
-	sub, err := a.contract.WatchChannelUpdate(watchOpts, events, [][32]byte{req.Params.ID()})
-	if err != nil {
-		err = cherrors.CheckIsChainNotReachableError(err)
-		return errors.WithMessage(err, "creating subscription failed")
+	defer sub.Close()
+	// Check for past event
+	if err := sub.ReadPast(ctx, events); err != nil {
```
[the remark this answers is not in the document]
ggwpez: Because i did not want to change the behaviour of the function.  
The function read the past events first, aborted if any, and then continued.  
This is different from just reading all events.  

It would surely be better if we could simplify all these functions, but i did not want to do that in this MR since it was out of scope.

backend/ethereum/channel/conclude.go
ggwpez: I do not want to block geth. Using un-buffered channels has no advantage here IMHO.

backend/ethereum/channel/conclude.go
ggwpez: Sadly not. We could add an extended filtering logic to the `EventSub`, i will record it on an issue.

backend/ethereum/channel/subscription.go
ggwpez: Sure it is possible to simplify this, but i did not want to blow up this MR by refactoring many functions.

backend/ethereum/channel/subscription.go
ggwpez: I dont know what this `RegisteredSub` is good for, i just replaced filter/watch with `EventSub`.

backend/ethereum/subscription/eventsub.go
ggwpez: This is not ideal, but i dont think it is possible to do better without the use of *reflections*.
`UnpackLog` does not set the `Raw` field of the `Event.Data` that it unpacks into.  
Since i do not know the exact type of `Event.Data`, the `Raw` field remains `nil`.  
This a bit of a trade-off between generic and concrete event subs.

backend/ethereum/channel/conclude.go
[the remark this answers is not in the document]
matthiasgeihs: I like your idea. So when we load the abi types, we also load/assert the event names.

backend/ethereum/channel/conclude.go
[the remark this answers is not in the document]
matthiasgeihs: We should focus on having a correct and clean implementation first, and then worry about how much it is consistent with the prior implementation.

The idea was that we can plug in our new subscription in most places without having to merge the past and future event streams manually. Are you saying that this is not possible?

backend/ethereum/channel/conclude.go
matthiasgeihs: Why not? Could it be because we are [not indexing the phase](https://github.com/hyperledger-labs/perun-eth-contracts/blob/27efac547588bc841151701000e289becf989739/contracts/Adjudicator.sol#L64)?

backend/ethereum/channel/subscription.go
matthiasgeihs: Hm, I don't see currently how this would blow up. You might want to explain during a meeting.

backend/ethereum/channel/subscription.go
matthiasgeihs: OK. As noted above, the idea was that we can replace the complicated existing filter/watch logic with our new sub and not worry about merging past and future events. This requires understanding the context.

backend/ethereum/channel/conclude.go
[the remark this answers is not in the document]
ggwpez: It is possible, but i am not comfortable with just changing the protocol i so many places without concrete concept.  
The past and future event stream merging is done here on purpose, since they are treated *differently*,

backend/ethereum/channel/conclude.go
```diff
@@ -35,31 +36,40 @@ const secondaryWaitBlocks = 2
...
+		return errors.WithMessage(err, "reading past events")
 	}
-	defer sub.Unsubscribe()
-
-	if found, err := a.filterConcluded(ctx, req.Params.ID()); err != nil {
-		return errors.WithMessage(err, "filtering old Concluded events")
-	} else if found {
-		return nil
+	select {
+	case _e := <-events:
```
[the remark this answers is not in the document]
matthiasgeihs: I think this does not work as intended. A concluded event is often preceded by a registered event. If you only check the first event, you will probably not get the concluded event.

backend/ethereum/channel/conclude.go
matthiasgeihs: OK

backend/ethereum/channel/conclude.go
```diff
@@ -35,31 +36,40 @@ const secondaryWaitBlocks = 2
 //   - if found, channel is already concluded and success is returned
 //   - if none found, conclude/concludeFinal is called on the adjudicator
 // - it waits for a Concluded event from the blockchain.
+// nolint: funlen
```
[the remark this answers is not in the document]
matthiasgeihs: For example, you could move the last for loop at the end to another function `waitConcluded`.

backend/ethereum/channel/subscription.go
matthiasgeihs: I made a suggestion.

backend/ethereum/channel/conclude.go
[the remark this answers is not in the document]
matthiasgeihs: OK, we can keep as is for now.

backend/ethereum/channel/conclude.go
```diff
@@ -69,24 +79,44 @@ func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.Adjudicat
 
 	// No conclude event found in the past, send transaction.
 	if req.Tx.IsFinal {
-		err = errors.WithMessage(a.callConcludeFinal(ctx, req), "calling concludeFinal")
+		err = cherrors.CheckIsChainNotReachableError(a.callConcludeFinal(ctx, req))
+		err = errors.WithMessage(err, "calling concludeFinal")
 	} else {
-		err = errors.WithMessage(a.callConclude(ctx, req, subStates), "calling conclude")
+		err = cherrors.CheckIsChainNotReachableError(a.callConclude(ctx, req, subStates))
```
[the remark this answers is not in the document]
matthiasgeihs: After discussion in meeting, we agreed to keep the logic similar to before. We may try to simplify later.

matthiasgeihs (changes requested):
Please address the linter warning. It's easy to fix. For one of them I already added a commit.

There is probably a bug in `ensureConcluded`. What if the first event is not the Concluded event?

backend/ethereum/channel/conclude.go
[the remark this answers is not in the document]
ggwpez: Right, i will replace it with a loop.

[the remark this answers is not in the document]
ggwpez: Added a `isConcluded` function now, which does the past querying.

backend/ethereum/channel/conclude.go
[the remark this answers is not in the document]
ggwpez: Moved the `ReadPast` into `isConcluded`.

backend/ethereum/channel/subscription.go
ggwpez: Looks good

matthiasgeihs (dismissed):
LGTM

matthiasgeihs (approved):

PR #96, opened 2021-06-07
Log message type.

### LINKED ISSUE #95: Log message type
**Location**: `[client]`

**Problem**: Having logging output like this is not useful:  
```pre
time="2021-06-07T11:59:58+02:00" level=debug msg="channelConn: publishing message: &{[65 118 88 117 50 30 136 25 23 181 229 133 193 244 89 48 252 220 160 8 68 111 236 240 215 1 236 233 187 133 123 215] 0 [236 88 2 227 250 155 243 158 46 97 169 75 86 125 210 9 34 189 195 213 195 28 222 82 123 62 78 20 133 132 5 235 89 165 161 71 31 229 119 245 179 117 90 231 243 144 55 62 172 210 79 237 137 157 9 175 98 165 218 90 209 210 248 167 54]}" 
```

**Solution**: Also log `msg.Type()`.

ggwpez:
Log the message type instead of a binary blob.

Closes #95

### DIFF
--- a/client/channelconn.go
+++ b/client/channelconn.go
@@ -99,7 +99,7 @@ func (c *channelConn) Send(ctx context.Context, msg wire.Msg) error {
 		if channel.Index(i) == c.idx {
 			continue // skip own peer
 		}
-		c.log.WithField("peer", peer).Debugf("channelConn: publishing message: %v", msg)
+		c.log.WithField("peer", peer).Debugf("channelConn: publishing message: %v: %+v", msg.Type(), msg)
 		env := &wire.Envelope{
 			Sender:    c.sender(),
 			Recipient: peer,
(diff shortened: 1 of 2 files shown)


matthiasgeihs (commented):
I agree that just printing the raw message data structure is not easily digestible for a human reader. However, sometimes you might still want more than just the type.

I suggest `Printf("%v: %v", msg.Type(), msg)`.

You may also consider:
- Implement the Stringer interface for `wire.Msg`
- Use `%T` for printing the type. E.g., `Printf("%T: %v", msg, msg)`
- Use `%+v` for more context on the data fields.

ggwpez:
> Implement the Stringer interface for `wire.Msg`

Since `wire.Msg` itself is an `interface`, (and golang does not have dynamic dispatching) I think the change would be too large.  

> I suggest Printf("%v: %v", msg.Type(), msg).  

Ok i will use `Printf("%v: %+v", msg.Type(), msg)` then, with the `%+v`.

> Use `%T` for printing the type. E.g., `Printf("%T: %v", msg, msg)`  

`%T` formats it a bit differently, eg: as `*client.msgChannelUpdateAcc` instead of `ChannelUpdateAcc`.

> Use `%+v` for more context on the data fields.  

Check

matthiasgeihs (approved):

PR #104, opened 2021-06-14
Add Auto-mining to SimulatedBackend

### LINKED ISSUE #103: Add Auto-mining to SimulatedBackend
**Location**: `[backend/eth/channel/test] simulated.go`

**Problem**: We need the ability to let the `SimulatedBackend` tick in order to test TX/Event finality.  
The `SimulatedBackend` is currently not producing blocks on its own, which means that the `finalityDepth` in the tests is never reached for a TX/Event.

**Solution**: Add a `SimulatedBackend.StartMining(interval time.Duration)` function which starts to produce an empty block every time that `interval` elapses.  
Add a `SimulatedBackend.StopMining` function which stops the mining and releases resources.  

The only restriction here is that the `blockTime` itself is fixed to 10s in *go-ethereum*. So everytime a new block is mined, it timestamp will be increased by 10s.  
I think this can not be prevented even with `AdjustTime` as that would race with other block creations.  
As a result of this, some tests need to be adjusted to use a higher `ChallengeDuration`, which seems to work out.

**Testing**: A simple test that checks that new blocks are created in between `StartMining` and `StopMining`.  
I implemented it for #97 and it seems to work fine when Incorporated into `backend/eth/channel/test.NewSetup`.

ggwpez:
Introduces two functions: `StartMining(interval)` and `StopMining`.  
Can be used to simulate a ticking blockchain.  

Closes #103

### DIFF
--- a/backend/ethereum/channel/test/simulated.go
+++ b/backend/ethereum/channel/test/simulated.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"math/big"
+	"time"
 
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind/backends"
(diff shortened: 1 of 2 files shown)


matthiasgeihs (approved):
looks good to me

PR #107, opened 2021-06-14
Remove hardcoded genesis alloc

### LINKED ISSUE #54: eth/channel/test: Replace custom genesis code in `SimulatedBackend` constructor with `core.DeveloperGenesisBlock`
Our [constructor](https://github.com/hyperledger-labs/go-perun/blob/25e0d303ebf6be4f658f3efa8d596512c0f49cff/backend/ethereum/channel/test/simulated.go#L47) still uses some, now outdated, copied code from [`core.DeveloperGenesisBlock`](https://github.com/ethereum/go-ethereum/blob/d8ff53dfb8a516f47db37dbc7fd7ad18a1e8a125/core/genesis.go#L399). We should use this function instead (passing it as faucet address the key that we generate at the top), avoiding code duplication and being automatically up to date with the latest pre-compiles.

I don't know the meaning of parameter `period`, that's up for research. Looks like it is relevant in our sim backend use case.

ggwpez:
I know that this has low prio but it was annoying me :laughing:  

Remove the hard coded genesis allocation addresses with a call to `DeceloperGenesisBlock`.

Closes #54

### DIFF
--- a/backend/ethereum/channel/test/simulated.go
+++ b/backend/ethereum/channel/test/simulated.go
@@ -27,7 +27,6 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/pkg/errors"
 
-	"perun.network/go-perun/channel"
 	"perun.network/go-perun/channel/test"
 	"perun.network/go-perun/pkg/sync"
 )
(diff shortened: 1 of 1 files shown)


ggwpez:
Will reopen when possible to fix.

PR #112, opened 2021-06-16
:sparkles: [pkg/test] Add context to ConcurrentT.

### LINKED ISSUE #110: [pkg/test] Improve ConcurrentT.Wait()
It would be very handy to extend `pkg/test/ConcurrentT` with a function `WaitCtx(ctx, stages...)` that fails and aborts the test if the context expires before the requested stages are done. This is important if we have contexts that should limit the duration of tests.

RmbRT:
ConcurrentT can now be constructed with an optional context which will cause all stages and waiting operations to fail on expiry.
I diverged from the issue's initial plan, because introducing custom contexts would have made the ConcurrentT object harder to use correctly. Now, there is only one context per ConcurrentT, and not one per wait call. The only adaption needed to use the new feature is to swap out the ConcurrentT's constructor in tests.

Closes #110.

### DIFF
--- a/pkg/test/concurrent.go
+++ b/pkg/test/concurrent.go
@@ -15,6 +15,7 @@
 package test
 
 import (
+	"context"
 	"runtime"
 	"strconv"
 	"strings"
@@ -130,6 +131,7 @@ type ConcurrentT struct {
 	t            require.TestingT
 	failed       bool
 	failedCh     chan struct{}
+	ctx          context.Context
 
 	mutex  sync.Mutex
 	stages map[string]*stage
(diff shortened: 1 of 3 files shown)


ggwpez (changes requested):
Thanks for adding the context errors via `t.Errorf`, but it seems that `Errorf` is not thread safe :man_shrugging:, see [CI](https://github.com/hyperledger-labs/go-perun/pull/112/checks?check_run_id=2848746965#step:4:159).  
@RmbRT

ggwpez (approved):
Great

Issue #105, opened 2021-06-14
Add Reorg to SimulatedBackend


ggwpez:
**Location**: `[backend/eth/channel/test] simulated.go`  

**Problem**: After the *go-ethereum* MR was merged in an altered version and the tests for #40 and #97 already use it, i think it makes sense to keep the `Reorg` function.  

**Note**: The miguelmota HD wallet does not work with the current *master* version of *go-ethereum*.  
We will probably have to do a replacement in the `go.mod` file to my fork of *go-etherum* since it has the `Fork` function but is based on an older release.  
This is only necessary until the new *go-ethereum* version is released and adapted in the HD wallet.

**Solution**: Use the new `Fork` function from the `SimulatedBackend` and add a `Rerog` function to our wrapper type.  

**Example**:
Assume we have two dependent transactions `A` and `B`. `B` can only be mined after `A` was mined.  
Each transaction is mined in its own block.  
The following code reorders `B` before `A` (with depth=2, length=3), therefore rendering `B` invalid. `A` remains valid.
```go
// Send A and B
// Confirm A (ok)
// Confirm B (ok)
sim.Reorg(ctx, 2, 3, func(txs []types.Transactions) []types.Transactions {
    return []types.Transactions{txs[1] /*B*/, txs[0] /*A*/, nil}
})
// Confirm A (ok)
// Confirm B (error)
```

**Tests**: Tests are quite complicated and long, not sure if we actually want to review/merge them…  
Will also be tested implicitly in #40 and #97.

PR #116, opened 2021-06-18
Resistant event subs

### LINKED ISSUE #40: Resistant Event Sub
**Location**: Package `[backend/eth/glue]`, part of #19

**Motivation**: The Resistant Event Sub tracks the state of an event in the case of a chain reorganization.  

**Design**: Analogous to #39 the design is very simple. It wraps an `EventSub` and accepts a `finalityDepth`. The necessary `ChainReader` is provided by the `ContractBackend`.  

It tracks the status of an Event by subscribing to new block chain headers and then checking all events for finality.  
Once an Event is final, it is emitted on the `sink` channel that the user passed to `Read`.

uml
[uml.txt](https://github.com/hyperledger-labs/go-perun/files/6280360/uml.txt)

**Testing**: So far i am testing that final Events arrive even after reorgs. Still needs to be slightly extended.

ggwpez:
Adds the `ResistantEventSub` type.  
Also introduced a convenience function `Subscribe(ctx context.Context, cr ethereum.ChainReader, contract *bind.BoundContract, eFact EventFactory, startBlockOffset, confirmations uint64) (*ResistantEventSub, error)` which makes it easier to create a resistant sub.  
I am not happy with the arguments of `Subscribe` yet, since it are so many.

Closes #40 
Draft until #105 is closed

### DIFF
--- a/backend/ethereum/subscription/resistanteventsub.go
+++ b/backend/ethereum/subscription/resistanteventsub.go
@@ -19,6 +19,7 @@ import (
 	"math/big"
 
 	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/pkg/errors"
@@ -44,6 +45,22 @@ type (
 		events       map[common.Hash]*Event
 	}
 )
+
+// Subscribe is a convenience function which returns a `ResistantEventSub`.
+// It is equivalent to manually calling `NewEventSub` and `NewResistantEventSub`
+// with the given parameters.
+func Subscribe(ctx context.Context, cr ethereum.ChainReader, contract *bind.BoundContract, eFact EventFactory, startBlockOffset, confirmations uint64) (*ResistantEventSub, error) {
+	_sub, err := NewEventSub(ctx, cr, contract, eFact, startBlockOffset)
+	if err != nil {
+		return nil, errors.WithMessage(err, "creating filter-watch event subscription")
+	}
+	sub, err := NewResistantEventSub(ctx, _sub, cr, confirmations)
+	if err != nil {
+		_sub.Close()
+		return nil, errors.WithMessage(err, "creating filter-watch event subscription")
+	}
+	return sub, nil
+}
 
 // NewResistantEventSub creates a new `ResistantEventSub` from the given
 // `EventSub`. Closes the passed `EventSub` when done.
--- a/backend/ethereum/subscription/resistanteventsub.go
+++ b/backend/ethereum/subscription/resistanteventsub.go
@@ -1,0 +1,232 @@
+// Copyright 2021 - See NOTICE file for copyright holders.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package subscription
+
+import (
+	"context"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/pkg/errors"
+
+	cherrors "perun.network/go-perun/backend/ethereum/channel/errors"
+	"perun.network/go-perun/log"
+	pkgsync "perun.network/go-perun/pkg/sync"
+)
+
+type (
+	// ResistantEventSub wraps an `EventSub` and makes it resistant to chain reorgs.
+	// It handles `removed` and `rebirth` events and has a `finalityDepth`
+	// threshold to decide when an Event is final.
+	// It will never emit the same event twice.
+	ResistantEventSub struct {
+		closer        pkgsync.Closer
+		sub           *EventSub
+		finalityDepth *big.Int
+
+		lastBlockNum *big.Int
+		heads        chan *types.Header
+		headSub      ethereum.Subscription
+		events       map[common.Hash]*Event
+	}
+)
+
+// NewResistantEventSub creates a new `ResistantEventSub` from the given
+// `EventSub`. Closes the passed `EventSub` when done.
+// `finalityDepth` defines in how many blocks an event needs to be included.
+// `finalityDepth` cannot be smaller than 1.
+// The passed `EventSub` should query more than `finalityDepth` blocks into
+// the past.
+func NewResistantEventSub(ctx context.Context, sub *EventSub, cr ethereum.ChainReader, finalityDepth uint64) (*ResistantEventSub, error) {
+	if finalityDepth < 1 {
+		panic("finalityDepth needs to be at least 1")
+	}
+	last, err := cr.HeaderByNumber(ctx, nil)
+	if err != nil {
+		err = cherrors.CheckIsChainNotReachableError(err)
+		return nil, errors.WithMessage(err, "subscribing to headers")
+	}
+	log.Debugf("Resistant Event sub started at block: %v", last.Number)
+	// Use a large buffer to not block geth.
+	heads := make(chan *types.Header, 128)
+	headSub, err := cr.SubscribeNewHead(ctx, heads)
+	if err != nil {
+		headSub.Unsubscribe()
+		err = cherrors.CheckIsChainNotReachableError(err)
+		return nil, errors.WithMessage(err, "subscribing to headers")
+	}
+
+	fd := new(big.Int)
+	fd.SetUint64(finalityDepth)
+	ret := &ResistantEventSub{
+		sub:           sub,
+		lastBlockNum:  last.Number,
+		heads:         heads,
+		headSub:       headSub,
+		finalityDepth: fd,
+		events:        make(map[common.Hash]*Event),
+	}
+	ret.closer.OnCloseAlways(func() {
+		headSub.Unsubscribe()
+		sub.Close()
+	})
+	return ret, nil
+}
+
+// Read reads all past and future events into `sink`.
+// Can be aborted by cancelling `ctx` or `Close()`.
+// All events can be considered final.
+func (s *ResistantEventSub) Read(_ctx context.Context, sink chan<- *Event) error {
+	ctx, cancel := context.WithCancel(_ctx)
+	defer cancel()
+	subErr := make(chan error, 1)
+	rawEvents := make(chan *Event, 128)
+	// Read events from the underlying event subscription.
+	go func() {
+		subErr <- s.sub.Read(ctx, rawEvents)
+	}()
+
+	for {
+		select {
+		case head := <-s.heads:
+			if head == nil {
+				return errors.New("head sub returned nil")
+			}
+			s.processHead(head, sink)
+		case event := <-rawEvents:
+			s.processEvent(event, sink)
+		case e := <-s.headSub.Err():
+			return errors.WithMessage(e, "underlying head subscription")
+		case e := <-subErr:
+			if e != nil {
+				return errors.WithMessage(e, "underlying EventSub.Read")
+			}
+			return errors.New("underlying event sub terminated")
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-s.closer.Closed():
+			return nil
+		}
+	}
+}
+
+// ReadPast reads all past events into `sink`.
+// Can be aborted by cancelling `ctx` or `Close()`.
+// All events can be considered final.
+func (s *ResistantEventSub) ReadPast(_ctx context.Context, sink chan<- *Event) error {
+	ctx, cancel := context.WithCancel(_ctx)
+	defer cancel()
+	subErr := make(chan error, 1)
+	rawEvents := make(chan *Event, 128)
+	// Read events from the underlying event subscription.
+	go func() {
+		defer close(rawEvents)
+		subErr <- s.sub.ReadPast(ctx, rawEvents)
+	}()
+
+	for {
+		select {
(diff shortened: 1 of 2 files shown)


backend/ethereum/subscription/eventsub.go
```diff
@@ -199,25 +205,25 @@ func (s *EventSub) readFuture(ctx context.Context, sink chan<- *Event) error {
...
-// Should be called exactly once and panics otherwise.
-// Must not be called if the construction function returned with an error.
+// Can be called more than once, but should be called at least once.
+// Is thread safe.
 func (s *EventSub) Close() {
-	close(s.closed)
-	s.watchSub.Unsubscribe()
-	s.filterSub.Unsubscribe()
+	if err := s.closer.Close(); !pkgsync.IsAlreadyClosedError(err) {
+		log.WithError(err).Error("could not close EventSub")
```
ggwpez, line 227: Regarding the error ignore here:  
I changed it to return the error but the change would propagate up to the usage of the `RegisteredSub` so i would rather do this in a separate issue.

backend/ethereum/subscription/eventsub.go
```diff
@@ -24,14 +24,16 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
```
ggwpez, line 24: Will fix the commit message after the first review.

backend/ethereum/subscription/resistanteventsub_test.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		log.Debug("Waiting for confirm")
+		OneEvent(require, sub)
+		sub.Close()
+		s.SB.Commit() // Make sure the next EventSub does not get some old block.
+	}
+}
+
+// TestResistantEventSub_FinalityDepth tests that a TX is confirmed exactly
+// after `finalityDepth` blocks even when reorgs occur.
```
matthiasgeihs, line 71:
```suggestion
// TestResistantEventSub_Reorg tests that a TX is confirmed exactly
// after `finalityDepth` blocks even when reorgs occur.
```

Also: Here you call it `finalityDepth`, above you call it confirmations. Either one or the other should be consistently used. I prefer `finalityDepth`.

backend/ethereum/subscription/resistanteventsub_test.go
```diff
@@ -0,0 +1,143 @@
...
+	require := require.New(t)
+	s := test.NewTokenSetup(ctx, t, rng)
+
+	for i := 0; i < 3; i++ {
+		finality := rng.Int31n(20) + 2
+		sub, err := subscription.Subscribe(ctx, s.CB, s.Contract, event, 0, uint64(finality))
+		require.NoError(err)
+
+		for i := 0; i < 20; i++ {
+			s.SB.Commit()
```
matthiasgeihs, line 85: 20 empty blocks? What is this supposed to simulate?

backend/ethereum/subscription/resistanteventsub_test.go
```diff
@@ -0,0 +1,143 @@
...
+	rng := pkgtest.Prng(t)
+	require := require.New(t)
+	s := test.NewTokenSetup(ctx, t, rng)
+
+	for i := 0; i < 3; i++ {
+		finality := rng.Int31n(20) + 2
+		sub, err := subscription.Subscribe(ctx, s.CB, s.Contract, event, 0, uint64(finality))
+		require.NoError(err)
+
+		for i := 0; i < 20; i++ {
```
matthiasgeihs, line 84: If this `20` here and the one above in line 80 refer to the same thing, you should use a const.

backend/ethereum/subscription/resistanteventsub_test.go
```diff
@@ -0,0 +1,143 @@
...
+			s.SB.Commit()
+		}
+		// Send and Confirm the TX
+		s.IncAllowance(ctx)
+		// Reorg until the block hight hits `finality` blocks.
+		for h := int64(0); h < int64(finality-1); {
+			d := rng.Int63n(10) + 1
+			l := rng.Int63n(10) + 1
```
matthiasgeihs, line 92: If the value 10 relates to the value 20 above in some way, I suggest to use a constant to ensure consistency in case one of the values is changed.

backend/ethereum/subscription/resistanteventsub_test.go
```diff
@@ -0,0 +1,143 @@
...
+			log.Debugf("[h=%d] Reorg with depth: %d, length: %d", h, d, l)
+			s.SB.Reorg(ctx, uint64(d), func(txs []types.Transactions) []types.Transactions {
+				return append(txs, make([]types.Transactions, int(d+l)-len(txs))...)
+			})
+			h += l
+		}
+		log.Debug("done")
+		// Verify that the event arrived.
+		OneEvent(require, sub)
+		sub.Close()
```
matthiasgeihs, line 103: Above, you add another `s.SB.Commit()` here. Looks like we don't need it? Or we should also add it here?

backend/ethereum/subscription/resistanteventsub_test.go
```diff
@@ -0,0 +1,143 @@
...
+	for i := 0; i < n; i++ {
+		require.NotNil(<-sink)
+	}
+
+	select {
+	case event := <-sink:
+		require.Nil(event)
+	default:
+	}
+}
```
matthiasgeihs, line 143: What do you think about adding a test for the case that an event is removed by a reorg and should therefore not be emitted?

backend/ethereum/subscription/resistanteventsub.go
```diff
@@ -0,0 +1,236 @@
...
+	if diff.Sign() > 0 {
+		for _, event := range s.events {
+			if s.isFinal(event) {
+				sink <- event
+				delete(s.events, event.Log.TxHash)
+			}
+		}
+	} else {
+		log.Debugf("Block number did not increase. Increment: %v", diff)
+	}
```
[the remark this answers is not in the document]
ggwpez: It is possible to just remove the check, the check just optimizes it slightly by only checking on increased block numbers.

backend/ethereum/subscription/eventsub.go
ggwpez: Move to #162

backend/ethereum/subscription/resistanteventsub.go
```diff
@@ -0,0 +1,236 @@
...
+		lastBlockNum *big.Int
+		heads        chan *types.Header
+		headSub      ethereum.Subscription
+		events       map[common.Hash]*Event
+	}
+)
+
+// NewResistantEventSub creates a new `ResistantEventSub` from the given
+// `EventSub`. Closes the passed `EventSub` when done.
+// `confirmations` defines in how many blocks the events needs to be included.
```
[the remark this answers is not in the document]
ggwpez: Yes will rename it, was annoyed from that as well.

backend/ethereum/subscription/resistanteventsub.go
```diff
@@ -0,0 +1,236 @@
...
+		case <-s.closer.Closed():
+			return nil
+		}
+	}
+}
+
+// ReadPast reads all past events into `sink`.
+// Can be aborted by cancelling `ctx` or `Close()`.
+// All events can be considered final.
+func (s *ResistantEventSub) ReadPast(_ctx context.Context, sink chan<- *Event) error {
```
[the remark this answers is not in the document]
ggwpez: Probably needed in the next MR, will check.

backend/ethereum/subscription/resistanteventsub.go
```diff
@@ -0,0 +1,236 @@
...
+				sink <- event
+				delete(s.events, event.Log.TxHash)
+			}
+		}
+	} else {
+		log.Debugf("Block number did not increase. Increment: %v", diff)
+	}
+}
+
+func (s *ResistantEventSub) isFinal(event *Event) bool {
```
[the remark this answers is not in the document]
ggwpez: Yes, things like that where the header and the event sub race each other can happen.  
https://github.com/hyperledger-labs/go-perun/issues/157 contains a similar case.

backend/ethereum/subscription/resistanteventsub_test.go
ggwpez: The blocks here are needed so that the following `Reorg` does not reorg blocks from the previous loop iteration.  
I think i will just remove the encompassing loop and only test it once.

backend/ethereum/subscription/resistanteventsub_test.go
ggwpez: I added a `maxFinality` constant.

backend/ethereum/subscription/resistanteventsub_test.go
ggwpez: Done

backend/ethereum/subscription/resistanteventsub_test.go
ggwpez: I remove the loops, so it is superfluous now.

backend/ethereum/subscription/resistanteventsub_test.go
ggwpez: Ok makes sense, analogous to the TX finality test.

ggwpez: Wow… adding such a test fails, since indeed, the header subscription seems to be much faster than the event subscription.

backend/ethereum/subscription/resistanteventsub.go
```diff
@@ -0,0 +1,236 @@
...
+		events:        make(map[common.Hash]*Event),
+	}
+	ret.closer.OnCloseAlways(func() {
+		headSub.Unsubscribe()
+		sub.Close()
+	})
+	return ret, nil
+}
+
+// Read reads all past and future events into `sink`.
```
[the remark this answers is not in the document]
ggwpez: We need all past event for example in the case of the `Funder`, which sums up all `Deposited` events.  
The `AdjSub` makes it easy to access the last `ChannelUpdate` event, where it does not really matter if you know the previous events or not.

backend/ethereum/subscription/resistanteventsub_test.go
ggwpez: I renamed it everywhere.

backend/ethereum/subscription/resistanteventsub.go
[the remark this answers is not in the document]
ggwpez: I added `TestResistantEventSub_ReadPast`, besides that it will be used in #117, .

backend/ethereum/subscription/resistanteventsub_test.go
```diff
@@ -0,0 +1,229 @@
...
+	require := require.New(t)
+	s := test.NewTokenSetup(ctx, t, rng)
+
+	finality := rng.Int31n(maxFinality-2) + 2
+	sub, err := subscription.Subscribe(ctx, s.CB, s.Contract, event, 0, uint64(finality))
+	require.NoError(err)
+	defer sub.Close()
+
+	// Send and Confirm the TX
+	s.ConfirmTx(s.IncAllowance(ctx), true)
```
matthiasgeihs, line 175: I have trouble following. Is it correct, that `ConfirmTx` waits until `TxFinalityDepth = 6` blocks until it returns? How is that the case here, if we do not progress the chain?

Update:
OK, checked in more detail. Realized that you set `TxFinalityDepth = 1` in the init function for this test. It doesn't seem like a good idea that we specify `TxFinalityDepth` as a global variable. It's hard to follow which value it currently has and there is also no protection against changing it on-the-fly, which may result in undefined behavior. Anyways, this is out of scope here. I will create an issue for that.

matthiasgeihs: Do we need to call `ConfirmTx` here at all?

backend/ethereum/subscription/resistanteventsub_test.go
```diff
@@ -0,0 +1,229 @@
...
+		return make([]types.Transactions, 2)
+	})
+
+	time.Sleep(1 * time.Second) // give geth time to update
+	NoEvent(require, sub)
+	// Verify that the event never arrives.
+	for i := 0; i < maxFinality; i++ {
+		s.SB.Commit()
+	}
+	time.Sleep(1 * time.Second) // give geth time to update
```
matthiasgeihs, line 189:
```suggestion
	time.Sleep(1 * time.Second) // give the event subscription time to catch up
```

backend/ethereum/subscription/resistanteventsub_test.go
matthiasgeihs: Looks like to increased that to a 1second wait.

.github/workflows/ci.yml
```diff
@@ -64,4 +64,4 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Unit Race Tests
-        run: go test -timeout 120s -race -gcflags=all=-d=checkptr=0 ./...
+        run: go test -timeout 240s -race -gcflags=all=-d=checkptr=0 ./...
```
matthiasgeihs, line 67: Not nice :( Does that mean our CI runs twice as long now?

Update:
I compared the running time of a CI run before and after the update. They are roughly equal. I don't think we need to bump the timeout, do we?

backend/ethereum/subscription/resistanteventsub_test.go
ggwpez: Right, i just added that yesterday for debugging, removed.

.github/workflows/ci.yml
ggwpez: The i dont know why it time-outed once… I will try it without again.

backend/ethereum/subscription/resistanteventsub_test.go
```diff
@@ -0,0 +1,229 @@
...
+	// Send and Confirm the TX
+	s.ConfirmTx(s.IncAllowance(ctx), true)
+	NoEvent(require, sub)
+
+	// Go back one block and remove the TX with a reorg.
+	s.SB.Reorg(ctx, 1, func(txs []types.Transactions) []types.Transactions {
+		return make([]types.Transactions, 2)
+	})
+
+	time.Sleep(1 * time.Second) // give geth time to update
```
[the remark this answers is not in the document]
ggwpez: I removed it.

matthiasgeihs (approved):

Issue #97, opened 2021-06-07
Resistant TX confirmation


ggwpez:
**Location**: `[backend/eth/channel] contractbackend.go`

**Problem**: TX are assumed to be final after one block, which is not the case for a reorg.

**Solution**: Implement a modified version of `ContractBackend.ConfirmTransaction` which checks that the Transaction is included in at least `TxBlockFinality` blocks.  
`TxBlockFinality` is a new package constant of `[backend/eth/channel]`. Value to be discussed in #98

PR #117, opened 2021-06-18
Integrate Resistant Event Subs

### LINKED ISSUE #85: Use resistant EventSubs in eth/channel
**Location**: `[backend/eth/channel]`  

**Motivation**: After #40 is done, the `ResistantEventSub` can be used in all places where `EventSub` is currently used to monitor events. This also helps to finalize the design of #40, which is currently in progress.  

**Testing**: Testing would take a bit of work, since the `SimulatedBackend` does not produce blocks on its own.  
A hacky PoW were a go-routine mines a block every 20ms did not work. I will look into to again, hopefully it can be done without much effort.  
Besides this, all simulated on-chain tests will need to be checked if they still work as intended.

ggwpez:
Uses resistant event confirmation on *go-perun* in all locations.  
Currently i am having trouble with the `AdjudicatorSub`, which seems to make some end-to-end tests fail.  
Funder and Adjudicator tests themselves seem to work.

Closes #85 
Draft until #40 and #97 are closed.

### DIFF
--- a/backend/ethereum/channel/conclude.go
+++ b/backend/ethereum/channel/conclude.go
@@ -59,10 +59,11 @@ func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.Adjudicat
 	}()
 
 	// In final Register calls, as the non-initiator, we optimistically wait for
-	// the other party to send the transaction first for secondaryWaitBlocks many
-	// blocks.
+	// the other party to send the transaction first for
+	// `secondaryWaitBlocks + TxFinalityDepth` many blocks.
 	if req.Tx.IsFinal && req.Secondary {
-		isConcluded, err := waitConcludedForNBlocks(waitCtx, a, events, secondaryWaitBlocks)
+		waitBlocks := secondaryWaitBlocks + int(TxFinalityDepth)
+		isConcluded, err := waitConcludedForNBlocks(waitCtx, a, events, waitBlocks)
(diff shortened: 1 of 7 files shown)


backend/ethereum/channel/conclude.go
```diff
@@ -28,7 +28,7 @@ import (
 	"perun.network/go-perun/channel"
 )
 
-const secondaryWaitBlocks = 2
+const secondaryWaitBlocks = 8
```
matthiasgeihs, line 31:
```suggestion
var secondaryWaitBlocks = TxFinalityDepth + 2
```

backend/ethereum/channel/conclude.go
```diff
@@ -28,7 +28,7 @@ import (
 	"perun.network/go-perun/channel"
 )
 
-const secondaryWaitBlocks = 2
+var secondaryWaitBlocks = int(TxFinalityDepth + 2)
```
matthiasgeihs, line 31: I just realized that my previous suggestion is problematic because `TxFinalityDepth` might change.

What do you think about keeping `const secondaryWaitBlocks = 2` here and when we use it in `conclude.go` L65, we write `waitConcludedForNBlocks(waitCtx, a, events, TxFinalityDepth + secondaryWaitBlocks)`?

matthiasgeihs (approved):

Issue #79, opened 2021-05-15
Indicate proposer index in channel proposal


matthiasgeihs:
Currently, in a channel proposal, we have the convention that the proposer has index 0. However, this is not documented well and not evident. Furthermore, the framework does not give an indication of that fact via, for example, a property on the proposal object (e.g., `proposal.ProposerIndex`) or the client package (e.g., `client.ProposerIndex`).

I think implicit assumptions like that are bad style. For example, if we, for whatever reason, decide to change away from this assumption, it will be difficult to make the framework user aware of that change. If we, at least, indicate this assumption via a property on some object, transitioning to a different behavior will be easier.

ggwpez:
Adding something like `client.ProposerIndex` seems reasonable to me.  
This could then also be used when creating the initial balance allocation or accessing a channel state.

sebastianst:
Yes, we followed this convention for simplicity and because in ledger channels it didn't seem like a limitation.

Exposing a `const client.ProposerIndex` constant sounds good.

It wouldn't be Go style to expose it as a static value on a struct, a package-wide constant would be better.

PR #123, opened 2021-06-24
Core82 open and close

matthiasgeihs:
Split some more commits from #83 .

In particular, this PR includes the following:

- Open virtual channels
- Close virtual channels
- Test virtual channels happy case

### DIFF
--- a/channel/test/randomizer.go
+++ b/channel/test/randomizer.go
@@ -222,6 +222,33 @@ func NewRandomChannelID(rng *rand.Rand, opts ...RandomOpt) (id channel.ID) {
 	return
 }
 
+// NewRandomChannelIDs generates a list of random channel IDs.
+func NewRandomChannelIDs(rng *rand.Rand, n int) (ids []channel.ID) {
+	ids = make([]channel.ID, n)
+	for i := range ids {
+		ids[i] = NewRandomChannelID(rng)
+	}
+	return
+}
+
+// NewRandomIndexMap generates a random index map.
+func NewRandomIndexMap(rng *rand.Rand, numParts int, numPartsParent int) (m []channel.Index) {
+	m = make([]channel.Index, numParts)
+	for i := range m {
+		m[i] = channel.Index(rng.Intn(numPartsParent))
+	}
+	return
+}
+
+// NewRandomIndexMaps generates a list of random index maps.
+func NewRandomIndexMaps(rng *rand.Rand, numParts int, numPartsParent int) (maps [][]channel.Index) {
+	maps = make([][]channel.Index, numParts)
+	for i := range maps {
+		maps[i] = NewRandomIndexMap(rng, numParts, numPartsParent)
+	}
+	return
+}
+
 // NewRandomBal generates a new random `channel.Bal`.
 // Options: `WithBalancesRange`.
 func NewRandomBal(rng *rand.Rand, opts ...RandomOpt) channel.Bal {
--- a/client/client.go
+++ b/client/client.go
@@ -34,16 +34,17 @@ import (
 //
 // Currently, only the two-party protocol is fully implemented.
 type Client struct {
-	address        wire.Address
-	conn           clientConn
-	channels       chanRegistry
-	funder         channel.Funder
-	adjudicator    channel.Adjudicator
-	wallet         wallet.Wallet
-	pr             persistence.PersistRestorer
-	log            log.Logger // structured logger for this client
-	version1Cache  version1Cache
-	fundingWatcher *stateWatcher
+	address           wire.Address
+	conn              clientConn
+	channels          chanRegistry
+	funder            channel.Funder
+	adjudicator       channel.Adjudicator
+	wallet            wallet.Wallet
+	pr                persistence.PersistRestorer
+	log               log.Logger // structured logger for this client
+	version1Cache     version1Cache
+	fundingWatcher    *stateWatcher
+	settlementWatcher *stateWatcher
 
 	sync.Closer
 }
@@ -103,6 +104,7 @@ func New(
 	}
 
 	c.fundingWatcher = newStateWatcher(c.matchFundingProposal)
+	c.settlementWatcher = newStateWatcher(c.matchSettlementProposal)
 	return
 }
 
@@ -172,6 +174,8 @@ func (c *Client) Handle(ph ProposalHandler, uh UpdateHandler) {
 		case *msgChannelUpdate:
 			go c.handleChannelUpdate(uh, env.Sender, msg)
 		case *virtualChannelFundingProposal:
+			go c.handleChannelUpdate(uh, env.Sender, msg)
+		case *virtualChannelSettlementProposal:
 			go c.handleChannelUpdate(uh, env.Sender, msg)
 		case *msgChannelSync:
 			go c.handleSyncMsg(env.Sender, msg)
--- a/client/client.go
+++ b/client/client.go
@@ -34,15 +34,16 @@ import (
 //
 // Currently, only the two-party protocol is fully implemented.
 type Client struct {
-	address       wire.Address
-	conn          clientConn
-	channels      chanRegistry
-	funder        channel.Funder
-	adjudicator   channel.Adjudicator
-	wallet        wallet.Wallet
-	pr            persistence.PersistRestorer
-	log           log.Logger // structured logger for this client
-	version1Cache version1Cache
+	address        wire.Address
+	conn           clientConn
+	channels       chanRegistry
+	funder         channel.Funder
+	adjudicator    channel.Adjudicator
+	wallet         wallet.Wallet
+	pr             persistence.PersistRestorer
+	log            log.Logger // structured logger for this client
+	version1Cache  version1Cache
+	fundingWatcher *stateWatcher
 
 	sync.Closer
 }
@@ -90,7 +91,7 @@ func New(
 		return nil, errors.WithMessage(err, "setting up client connection")
 	}
 
-	return &Client{
+	c = &Client{
 		address:     address,
 		conn:        conn,
 		channels:    makeChanRegistry(),
@@ -99,7 +100,10 @@ func New(
 		wallet:      wallet,
 		pr:          persistence.NonPersistRestorer,
 		log:         log,
-	}, nil
+	}
+
+	c.fundingWatcher = newStateWatcher(c.matchFundingProposal)
+	return
 }
 
 // Close closes this state channel client.
@@ -158,15 +162,19 @@ func (c *Client) Handle(ph ProposalHandler, uh UpdateHandler) {
 		}
 		msg := env.Msg
 
-		switch msg.Type() {
-		case wire.LedgerChannelProposal:
-			go c.handleChannelProposal(ph, env.Sender, msg.(*LedgerChannelProposal))
-		case wire.SubChannelProposal:
-			go c.handleChannelProposal(ph, env.Sender, msg.(*SubChannelProposal))
-		case wire.ChannelUpdate:
-			go c.handleChannelUpdate(uh, env.Sender, msg.(*msgChannelUpdate))
-		case wire.ChannelSync:
-			go c.handleSyncMsg(env.Sender, msg.(*msgChannelSync))
+		switch msg := msg.(type) {
+		case *LedgerChannelProposal:
+			go c.handleChannelProposal(ph, env.Sender, msg)
+		case *SubChannelProposal:
+			go c.handleChannelProposal(ph, env.Sender, msg)
+		case *VirtualChannelProposal:
+			go c.handleChannelProposal(ph, env.Sender, msg)
+		case *msgChannelUpdate:
+			go c.handleChannelUpdate(uh, env.Sender, msg)
+		case *virtualChannelFundingProposal:
+			go c.handleChannelUpdate(uh, env.Sender, msg)
+		case *msgChannelSync:
+			go c.handleSyncMsg(env.Sender, msg)
 		default:
 			c.log.Error("Unexpected %T message received in request loop")
 		}
--- a/client/client_persistence_test.go
+++ b/client/client_persistence_test.go
@@ -15,48 +15,22 @@
 package client_test
 
 import (
-	"math/big"
 	"math/rand"
 	"testing"
 
-	"perun.network/go-perun/apps/payment"
 	chprtest "perun.network/go-perun/channel/persistence/test"
-	chtest "perun.network/go-perun/channel/test"
-	"perun.network/go-perun/client"
 	ctest "perun.network/go-perun/client/test"
-	"perun.network/go-perun/pkg/test"
-	"perun.network/go-perun/wire"
 )
 
 func TestPersistencePetraRobert(t *testing.T) {
-	rng := test.Prng(t)
-	for i := 0; i < 2; i++ {
-		setups := NewSetupsPersistence(t, rng, []string{"Petra", "Robert"})
-		roles := [2]ctest.Executer{
+	runTwoPartyTest(t, func(rng *rand.Rand) (setups []ctest.RoleSetup, roles [2]ctest.Executer) {
+		setups = NewSetupsPersistence(t, rng, []string{"Petra", "Robert"})
+		roles = [2]ctest.Executer{
 			ctest.NewPetra(setups[0], t),
 			ctest.NewRobert(setups[1], t),
 		}
-
-		app := client.WithoutApp()
-		if i == 1 {
-			app = client.WithApp(
-				chtest.NewRandomAppAndData(rng, chtest.WithAppRandomizer(new(payment.Randomizer))),
-			)
-		}
-
-		cfg := &ctest.PetraRobertExecConfig{
-			BaseExecConfig: ctest.MakeBaseExecConfig(
-				[2]wire.Address{setups[0].Identity.Address(), setups[1].Identity.Address()},
-				chtest.NewRandomAsset(rng),
-				[2]*big.Int{big.NewInt(100), big.NewInt(100)},
-				app,
-			),
-			NumPayments: [2]int{2, 2},
-			TxAmounts:   [2]*big.Int{big.NewInt(5), big.NewInt(3)},
-		}
-
-		ctest.ExecuteTwoPartyTest(t, roles, cfg)
-	}
+		return
+	})
 }
 
 func NewSetupsPersistence(t *testing.T, rng *rand.Rand, names []string) []ctest.RoleSetup {
--- a/client/client_role_test.go
+++ b/client/client_role_test.go
@@ -27,14 +27,14 @@ import (
 	ctest "perun.network/go-perun/client/test"
 	"perun.network/go-perun/log"
 	wtest "perun.network/go-perun/wallet/test"
-	"perun.network/go-perun/wire"
+	wiretest "perun.network/go-perun/wire/test"
 )
 
 const roleOperationTimeout = 1 * time.Second
 
 func NewSetups(rng *rand.Rand, names []string) []ctest.RoleSetup {
 	var (
-		bus   = wire.NewLocalBus()
+		bus   = wiretest.NewSerializingLocalBus()
 		n     = len(names)
 		setup = make([]ctest.RoleSetup, n)
 	)
--- a/client/client_role_test.go
+++ b/client/client_role_test.go
@@ -18,9 +18,12 @@ import (
 	"context"
 	"math/rand"
 	"sync"
+	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"perun.network/go-perun/channel"
+	"perun.network/go-perun/client"
 	ctest "perun.network/go-perun/client/test"
 	"perun.network/go-perun/log"
 	wtest "perun.network/go-perun/wallet/test"
@@ -60,6 +63,26 @@ func NewSetups(rng *rand.Rand, names []string) []ctest.RoleSetup {
 	}
 
 	return setup
+}
+
+type Client struct {
+	*client.Client
+	ctest.RoleSetup
+}
+
+func NewClients(rng *rand.Rand, names []string, t *testing.T) []*Client {
+	setups := NewSetups(rng, names)
+	clients := make([]*Client, len(setups))
+	for i, setup := range setups {
+		setup.Identity = setup.Wallet.NewRandomAccount(rng)
+		cl, err := client.New(setup.Identity.Address(), setup.Bus, setup.Funder, setup.Adjudicator, setup.Wallet)
+		assert.NoError(t, err)
+		clients[i] = &Client{
+			Client:    cl,
+			RoleSetup: setup,
+		}
+	}
+	return clients
 }
 
 type (
--- a/client/proposal.go
+++ b/client/proposal.go
@@ -39,6 +39,7 @@ type (
 	ProposalHandler interface {
 		// HandleProposal is the user callback called by the Client on an incoming channel
 		// proposal.
+		// The response on the proposal responder must be called within the same go routine.
 		HandleProposal(ChannelProposal, *ProposalResponder)
 	}
 
@@ -147,6 +148,13 @@ func (c *Client) ProposeChannel(ctx context.Context, prop ChannelProposal) (*Cha
 		c.log.Panic("invalid nil argument")
 	}
 
+	// Prepare and cleanup, e.g., for locking and unlocking parent channel.
+	err := c.prepareChannelOpening(ctx, prop, proposerIdx)
+	if err != nil {
+		return nil, errors.WithMessage(err, "preparing channel opening")
+	}
+	defer c.cleanupChannelOpening(prop, proposerIdx)
+
 	// 1. validate input
 	peer := c.proposalPeers(prop)[proposeeIdx]
 	if err := c.validTwoPartyProposal(prop, proposerIdx, peer); err != nil {
@@ -166,14 +174,47 @@ func (c *Client) ProposeChannel(ctx context.Context, prop ChannelProposal) (*Cha
 	return ch, fundingErr
 }
 
+func (c *Client) prepareChannelOpening(ctx context.Context, prop ChannelProposal, ourIdx channel.Index) (err error) {
+	_, parentCh, err := c.proposalParent(prop, ourIdx)
+	if err != nil {
+		return
+	}
+	if parentCh != nil {
+		if !parentCh.machMtx.TryLockCtx(ctx) {
+			return ctx.Err()
+		}
+	}
+	return
+}
+
+func (c *Client) cleanupChannelOpening(prop ChannelProposal, ourIdx channel.Index) {
+	_, parentCh, err := c.proposalParent(prop, ourIdx)
+	if err != nil {
+		c.log.Warn("getting proposal parent:", err)
+		return
+	}
+	if parentCh != nil {
+		parentCh.machMtx.Unlock()
+	}
+}
+
 // handleChannelProposal implements the receiving side of the (currently)
 // two-party channel proposal protocol.
 // The proposer is expected to be the first peer in the participant list.
 //
 // This handler is dispatched from the Client.Handle routine.
-func (c *Client) handleChannelProposal(
-	handler ProposalHandler, p wire.Address, req ChannelProposal) {
-	if err := c.validTwoPartyProposal(req, proposeeIdx, p); err != nil {
+func (c *Client) handleChannelProposal(handler ProposalHandler, p wire.Address, req ChannelProposal) {
+	ourIdx := channel.Index(proposeeIdx)
+
+	// Prepare and cleanup, e.g., for locking and unlocking parent channel.
+	err := c.prepareChannelOpening(c.Ctx(), req, ourIdx)
+	if err != nil {
+		c.log.Warn("preparing channel opening:", err)
+		return
+	}
+	defer c.cleanupChannelOpening(req, ourIdx)
+
+	if err := c.validTwoPartyProposal(req, ourIdx, p); err != nil {
 		c.logPeer(p).Debugf("received invalid channel proposal: %v", err)
 		return
 	}
@@ -301,7 +342,7 @@ func (c *Client) proposeTwoPartyChannel(
 // checked.
 func (c *Client) validTwoPartyProposal(
 	proposal ChannelProposal,
-	ourIdx int,
+	ourIdx channel.Index,
 	peerAddr wallet.Address,
 ) error {
 	if err := proposal.Valid(); err != nil {
@@ -317,6 +358,10 @@ func (c *Client) validTwoPartyProposal(
 		return errors.Errorf("expected 2 peers, got %d", len(peers))
 	}
 
+	if !(ourIdx == proposerIdx || ourIdx == proposeeIdx) {
+		return errors.Errorf("invalid index: %d", ourIdx)
+	}
+
 	peerIdx := ourIdx ^ 1
 	// In the 2PCPP, the proposer is expected to have index 0
 	if !peers[peerIdx].Equals(peerAddr) {
@@ -328,8 +373,13 @@ func (c *Client) validTwoPartyProposal(
 		return errors.Errorf("we don't have peer index %d", ourIdx)
 	}
 
-	if proposal.Type() == wire.SubChannelProposal {
-		if err := c.validSubChannelProposal(proposal.(*SubChannelProposal)); err != nil {
+	switch prop := proposal.(type) {
+	case *SubChannelProposal:
+		if err := c.validSubChannelProposal(prop); err != nil {
+			return errors.WithMessage(err, "validate subchannel proposal")
+		}
+	case *VirtualChannelProposal:
+		if err := c.validVirtualChannelProposal(prop, ourIdx); err != nil {
 			return errors.WithMessage(err, "validate subchannel proposal")
 		}
 	}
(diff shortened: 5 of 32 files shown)


client/client_role_test.go
```diff
@@ -18,9 +18,12 @@ import (
 	"context"
 	"math/rand"
 	"sync"
+	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
```
ggwpez, line 24: I could probably write a CI script which checks the grouping automatically :laughing:

ggwpez (changes requested):
Three things that came to my mind:  
1. There seem to be some tests missing for `virtualChannelSettlementProposal.En/Decode` and `virtualChannelFundingProposal.En/Decode`.
2. I think we should add the code-cov bot that we also have in the contracts repo to see if there is untested code.
3. The file naming is now inconsistent (eg. `virtual_channel_util.go`) since we normally only use an underscore for `_test` files. What do you think @matthiasgeihs?

client/proposal.go
```diff
@@ -317,6 +317,10 @@ func (c *Client) validTwoPartyProposal(
 		return errors.Errorf("expected 2 peers, got %d", len(peers))
 	}
 
+	if !(ourIdx == proposerIdx || ourIdx == proposeeIdx) {
```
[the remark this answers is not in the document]
matthiasgeihs: Yes, we should consider that. There is already an issue for that, #79 .

client/proposal.go
```diff
@@ -356,6 +365,38 @@ func (c *Client) validSubChannelProposal(proposal *SubChannelProposal) error {
...
+	parent, err := c.Channel(prop.Parents[ourIdx])
+	if err != nil {
+		return errors.New("parent channel not found")
+	}
+
+	if err := channel.AssetsAssertEqual(parent.State().Assets, prop.InitBals.Assets); err != nil {
+		return errors.WithMessage(err, "unequal assets")
+	}
+
+	if !prop.InitBals.Balances.Equal(prop.FundingAgreement) {
```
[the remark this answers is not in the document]
matthiasgeihs: Not sure if it makes sense in the context of virtual channels. Actually, this is one of the features that I don't see an immediate use case for while it adds complexity to the code base and also makes the library usage more complicated.

client/proposal.go
```diff
@@ -356,6 +365,38 @@ func (c *Client) validSubChannelProposal(proposal *SubChannelProposal) error {
...
+		return errors.WithMessage(err, "unequal funding agreement")
+	}
+
+	numIndexMaps := len(prop.IndexMaps)
+	if numIndexMaps != numPeers {
+		return errors.Errorf("expected %d index maps, got %d", numPeers, numIndexMaps)
+	}
+
```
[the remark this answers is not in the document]
matthiasgeihs: No, this would translate the parent balances, but here we want to translate the proposed balances.

client/proposal.go
```diff
@@ -356,6 +365,38 @@ func (c *Client) validSubChannelProposal(proposal *SubChannelProposal) error {
...
+	virtualBals := transformBalances(prop.InitBals.Balances, parent.state().NumParts(), prop.IndexMaps[ourIdx])
```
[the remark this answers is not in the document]
matthiasgeihs: Thanks for catching this one. The parent channel is not locked, but you are right that it would be better if it was. However, ensuring this is a bit tricky. I think we should have call about this.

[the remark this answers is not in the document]
matthiasgeihs: Fixed, please have a look.

client/proposal.go
```diff
@@ -469,6 +505,24 @@ func (c *Client) completeCPP(
...
 
+func (c *Client) proposalParent(prop ChannelProposal, partIdx channel.Index) (parentChannelID *channel.ID, parent *Channel, err error) {
+	switch prop := prop.(type) {
+	case *SubChannelProposal:
+		parentChannelID = &prop.Parent
+	case *VirtualChannelProposal:
+		parentChannelID = &prop.Parents[partIdx]
+	}
+
+	if parentChannelID != nil {
```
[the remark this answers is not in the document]
matthiasgeihs: This is intentional. In this case `parentChannelID` is `nil`.

sebastianst (commented):
Partial review of first commit.

client/client.go
```diff
@@ -106,6 +107,10 @@ func New(
 		entries:   make(map[uint]watcherEntry),
 		condition: c.matchFundingProposal,
 	}
+	c.settlementWatcher = &stateWatcher{
+		entries:   make(map[uint]watcherEntry),
+		condition: c.matchSettlementProposal,
+	}
```
sebastianst: Please add `newStateWatcher(condition) *stateWatcher` constructor to avoid repetition of internals (making a map).

client/subchannel.go
```diff
@@ -29,7 +29,22 @@ func (c *Channel) IsLedgerChannel() bool {
 
 // IsSubChannel returns whether the channel is a sub-channel.
 func (c *Channel) IsSubChannel() bool {
-	return c.Parent() != nil
+	return c.Parent() != nil && c.equalParticipants(c.Parent())
```
sebastianst, line 32: Can't we actually read this from the params?

client/updatemsgs.go
```diff
@@ -191,14 +191,22 @@ Virtual channel
...
 		InitialStateSignatures []wallet.Sig
 		IndexMap               []channel.Index
 	}
+
+	// virtualChannelSettlementProposal is a channel update that proposes the settlement of a virtual channel.
+	virtualChannelSettlementProposal struct {
+		msgChannelUpdate
+		ChannelParams        channel.Params
+		FinalState           channel.State
+		FinalStateSignatures []wallet.Sig
```
sebastianst, line 208: Why not use `Final SignedState` instead?

client/updatemsgs.go
```diff
@@ -223,3 +231,24 @@ func (m *virtualChannelFundingProposal) Decode(r io.Reader) (err error) {
...
+	)
+}
+
+func (m *virtualChannelSettlementProposal) Decode(r io.Reader) (err error) {
+	if err := perunio.Decode(r,
+		&m.msgChannelUpdate,
+		&m.ChannelParams,
+		&m.FinalState,
+		&m.FinalStateSignatures,
+	); err != nil {
```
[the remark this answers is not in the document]
sebastianst: just return in the first line `return perunio.Decode(...`

client/updatemsgs.go
```diff
@@ -223,3 +231,24 @@ func (m *virtualChannelFundingProposal) Decode(r io.Reader) (err error) {
 	}
 	return err
 }
+
+func (m virtualChannelSettlementProposal) Encode(w io.Writer) error {
+	return perunio.Encode(w,
+		m.msgChannelUpdate,
+		m.ChannelParams,
```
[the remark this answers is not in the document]
sebastianst: yes, also please test this with the generic serialization test.

client/virtual_channel_settlement.go
```diff
@@ -0,0 +1,188 @@
...
+	parent *Channel,
+	prop *virtualChannelSettlementProposal,
+	responder *UpdateResponder,
+) {
+	err := c.validateVirtualChannelSettlementProposal(parent, prop)
+	if err != nil {
+		c.rejectProposal(responder, err.Error())
+	}
+
+	ctx, cancel := context.WithTimeout(c.Ctx(), virtualChannelFundingTimeout)
```
sebastianst, line 77: maybe rename `virtualChannelFundingTimeout` to something like `virtualChannelUpdateTimeout` because it is not only used for funding now.

client/virtual_channel_settlement.go
```diff
@@ -0,0 +1,188 @@
...
+	// Cast.
+	inputs := []interface{}{a, b}
+	props := make([]*virtualChannelSettlementProposal, len(inputs))
+	for i, x := range inputs {
+		var prop *virtualChannelSettlementProposal
+		prop, ok = x.(*virtualChannelSettlementProposal)
+		if !ok {
+			return
+		}
+		props[i] = prop
```
sebastianst, line 151: can shorten again and just set the slice vals directly

client/virtual_channel_settlement.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	prop0 := props[0]
+
+	for _, prop := range props {
+		// Check final state.
+		ok = prop.FinalState.Equal(&prop0.FinalState) == nil
+		if !ok {
+			return
+		}
```
sebastianst, line 161: Again, the use of `ok` is a bit confusing here.

client/virtual_channel_settlement.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	// Store settlement state and signature.
+	virtual, err := c.Channel(prop0.FinalState.ID)
+	if err != nil {
+		return false
+	}
+	if err = virtual.machine.Update(c.Ctx(), &prop0.FinalState, hubIndex); err != nil {
```
sebastianst, line 169: Does the state machine allow the skipping of version numbers? Did you test to do a couple of direct updates in the virtual channels so that the version is at least 2 when settling?

client/virtual_channel_settlement.go
```diff
@@ -0,0 +1,188 @@
...
+	if err = virtual.machine.EnableFinal(c.Ctx()); err != nil {
+		return false
+	}
+
+	// Close channel.
+	err = virtual.Close()
+	if err != nil {
+		return false
+	}
+	c.channels.Delete(virtual.ID())
```
sebastianst, line 186: Not sure, but don't we also want to remove the virtual channel from persistence now? Or maybe I'm missing that this is happening somewhere else already...

client/virtual_channel_settlement.go
```diff
@@ -0,0 +1,188 @@
...
+
+	for _, prop := range props {
+		// Check final state.
+		ok = prop.FinalState.Equal(&prop0.FinalState) == nil
+		if !ok {
+			return
+		}
+	}
+
+	// Store settlement state and signature.
```
sebastianst, line 164: fix doc

client/virtual_channel_settlement.go
```diff
@@ -0,0 +1,188 @@
...
+	correctBalances := parent.state().Balances.Add(virtual).Equal(prop.State.Balances)
+	if !correctBalances {
+		return errors.New("invalid balances")
+	}
+
+	return nil
+}
+
+func (c *Client) matchSettlementProposal(a, b interface{}) (ok bool) {
+	// Cast.
```
sebastianst, line 142: Everywhere where we `return false` (or just after line 155), it might be worth logging with `c.Log().Warnf`, because it either indicates cheating peers or some bug, right? Right now, we just `return false` and the problems or `err` errors go unnoticed.

sebastianst: I think this also applies to the other matcher function

sebastianst: You could also do something like
```go
var err error
defer func() {
  if ok == false {
    c.Log().Warnf("Error in matching settlement proposal: %+v", err)
  }
}()
```
and then set the `err` variable to a descriptive error if we don't get one already from a function call.

sebastianst (commented):
2nd partial review of 2nd commit.

client/proposalmsgs.go
```diff
@@ -543,3 +561,110 @@ func (rej ChannelProposalRej) Encode(w io.Writer) error {
...
+		Peers:               peers,
+		Parents:             parents,
+		IndexMaps:           indexMaps,
+	}
+	return
+}
+
+// Encode encodes the proposal into an io.Writer.
+func (p VirtualChannelProposal) Encode(w io.Writer) error {
+	return perunio.Encode(w, p.Proposer, p.Peers, p.Parents, p.IndexMaps)
```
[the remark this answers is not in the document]
matthiasgeihs: Turns out, no. This was just not covered in the tests. I thought the encoding is done during the end-to-end test automatically, but turns out that the local bus does not encode/decode.

client/proposalmsgs.go
```diff
@@ -543,3 +561,110 @@ func (rej ChannelProposalRej) Encode(w io.Writer) error {
...
+}
+
+// Decode decodes a proposal from an io.Reader.
+func (p *VirtualChannelProposal) Decode(r io.Reader) error {
+	return perunio.Decode(r, &p.BaseChannelProposal, &p.Proposer, &p.Peers, &p.Parents, &p.IndexMaps)
```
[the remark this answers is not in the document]
matthiasgeihs: My bad, was untested, see comment above.

client/updatemsgs.go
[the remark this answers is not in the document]
matthiasgeihs: it didn't work, was untested, see comments above, added tests.

client/subchannel.go
matthiasgeihs: You mean by a flag? Unfortunately not. The original idea was that the params don't contain that information and they still don't, even though they now contain flags for virtualChannel and ledgerChannel.

client/virtual_channel.go
```diff
@@ -25,6 +25,13 @@ import (
 	"perun.network/go-perun/wire"
 )
 
+// IsVirtualChannel returns whether the channel is a virtual channel.
+// A virtual channel is a channel that has a parent channel with different
+// participants.
+func (c *Channel) IsVirtualChannel() bool {
+	return c.Parent() != nil && !c.equalParticipants(c.Parent())
```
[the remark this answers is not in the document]
matthiasgeihs: Currently, yes, but the flag is supposed to be there only temporary until we accept virtual channels in virtual channels.

client/virtual_channel_settlement.go
matthiasgeihs: Couldn't find a good name for both cases, added a separate variable for settling.

client/virtual_channel_settlement.go
matthiasgeihs: Not sure what you mean. Did you explain somewhere?

client/virtual_channel_settlement.go
matthiasgeihs: Hm?

client/virtual_channel_settlement.go
matthiasgeihs: Nice catch. I only did exactly one update in the test 🙈 Now added an additional transaction to the test and a ForceUpdate method.

client/virtual_channel_settlement.go
matthiasgeihs: we still need the state to register the channel if the update fails, right?

client/client_role_test.go
matthiasgeihs: It's not really that important I think.

client/virtual_channel_happy_test.go
```diff
@@ -0,0 +1,203 @@
...
+		rng,
+		[]string{"Alice", "Bob", "Ingrid"},
+		t,
+	)
+	alice, bob, ingrid := clients[0], clients[1], clients[2]
+
+	proposalHandlerIngrid := &FunctionProposalHandler{
+		openingProposalHandler: func(cp client.ChannelProposal, pr *client.ProposalResponder) {
+			switch cp := cp.(type) {
+			case *client.LedgerChannelProposal:
```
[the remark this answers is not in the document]
matthiasgeihs: Yes, but I rather do it programatically.

client/virtual_channel_happy_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (h *FunctionProposalHandler) HandleProposal(p client.ChannelProposal, r *client.ProposalResponder) {
+	h.openingProposalHandler(p, r)
+}
+
+func (h *FunctionProposalHandler) HandleUpdate(prev *channel.State, next client.ChannelUpdate, r *client.UpdateResponder) {
+	h.updateProposalHandler(prev, next, r)
+}
+
+func TestVirtualChannelsOptimistic(t *testing.T) {
```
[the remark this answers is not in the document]
matthiasgeihs: Definitely, I already did that when I added the dispute tests, which will be in one of the next PRs.

client/update.go
```diff
@@ -278,6 +278,7 @@ func (c *Channel) handleUpdateReq(
 	pidx channel.Index,
 	req ChannelUpdateProposal,
 	uh UpdateHandler,
+	client *Client,
```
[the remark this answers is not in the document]
matthiasgeihs: Already did that afterwards because it was necessary at some other point. Using it here now as well.

client/virtual_channel.go
```diff
@@ -0,0 +1,293 @@
...
+	parent, ok := c.channels.Get(parentID)
+	if !ok {
+		return errors.New("referenced parent channel not found")
+	}
+
+	indexMap := prop.IndexMaps[virtual.Idx()]
+	err = parent.proposeVirtualChannelFunding(ctx, virtual, indexMap)
+	if err != nil {
+		return errors.WithMessage(err, "proposing channel funding")
+	}
```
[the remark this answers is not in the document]
matthiasgeihs: I think readability suffers when pulling the call into the if. Named variables also help with readability, I feel, because when you have multiple return values, you do not need state all of them at every return statement.

client/virtual_channel.go
```diff
@@ -0,0 +1,293 @@
...
+
+	state := c.state().Clone()
+	state.Version++
+
+	// Deposit initial balances into sub-allocation
+	balances := virtual.translateBalances(indexMap)
+	state.Allocation.Balances = state.Allocation.Balances.Sub(balances)
+	state.AddSubAlloc(*channel.NewSubAlloc(virtual.ID(), balances.Sum(), indexMap))
+
+	err = c.updateGeneric(ctx, state, func(mcu *msgChannelUpdate) wire.Msg {
```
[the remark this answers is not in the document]
matthiasgeihs: I think it's semantically more clear to follow the same pattern. When a return call is followed by a rather long function call, it doesn't read nicely.

client/virtual_channel.go
```diff
@@ -0,0 +1,293 @@
...
+	err = ch.machine.SetFunded(ctx)
+	if err != nil {
+		return
+	}
+
+	if err := c.pr.ChannelCreated(ctx, ch.machine, peers, nil); err != nil {
+		return ch, errors.WithMessage(err, "persisting new channel")
+	}
+	ok := c.channels.Put(cID, ch)
+	if !ok {
```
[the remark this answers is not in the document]
matthiasgeihs: As explained above, I prefer it the way I did it, because I think it aids readability.

client/virtual_channel.go
```diff
@@ -0,0 +1,293 @@
...
+	// Cast.
+	inputs := []interface{}{a, b}
+	props := make([]*virtualChannelFundingProposal, len(inputs))
+	for i, x := range inputs {
+		var prop *virtualChannelFundingProposal
+		prop, ok = x.(*virtualChannelFundingProposal)
+		if !ok {
+			return
+		}
+		props[i] = prop
```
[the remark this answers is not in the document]
matthiasgeihs: Outdated now.

client/virtual_channel.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	prop0 := props[0]
+
+	// Check initial state.
+	for _, prop := range props {
+		ok = prop.InitialState.Equal(&prop0.InitialState) == nil
+		if !ok {
+			return
+		}
```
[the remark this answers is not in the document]
matthiasgeihs: I don't agree, as explained above. We can have a discussion offline, but changing it consistently across the whole code, I don't see much benefit in and it's a lot of additional work.

client/virtual_channel.go
```diff
@@ -0,0 +1,293 @@
...
+	// Deposit initial balances into sub-allocation
+	balances := virtual.translateBalances(indexMap)
+	state.Allocation.Balances = state.Allocation.Balances.Sub(balances)
+	state.AddSubAlloc(*channel.NewSubAlloc(virtual.ID(), balances.Sum(), indexMap))
+
+	err = c.updateGeneric(ctx, state, func(mcu *msgChannelUpdate) wire.Msg {
+		return &virtualChannelFundingProposal{
+			msgChannelUpdate:       *mcu,
+			ChannelParams:          *virtual.Params(),
+			InitialState:           *virtual.State(),
```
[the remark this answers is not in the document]
matthiasgeihs: Here it doesn't matter because we are still in the setup of the channel and it is not used yet. Hence, I am using the default, `Channel.State()`.

client/virtual_channel.go
```diff
@@ -0,0 +1,293 @@
...
+			InitialStateSignatures: virtual.machine.CurrentTX().Sigs,
```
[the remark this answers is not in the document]
matthiasgeihs: Changed to `Initial SignedState`.

client/virtual_channel_util.go
```diff
@@ -0,0 +1,133 @@
...
+
+type watcherEntry struct {
+	state interface{}
+	done  chan struct{}
+}
+
+type stateWatcher struct {
+	sync.RWMutex
+	entries   map[uint]watcherEntry
+	counter   uint
```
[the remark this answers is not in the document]
matthiasgeihs: I was considering that, but it has some downsides: https://forum.golangbridge.org/t/using-interface-types-as-map-keys/706

In particular, it would mean we have to be careful what we use the watcher for. Otherwise it might panic.

client/virtual_channel_settlement.go
matthiasgeihs: OK, only now saw the first part of your review. See comments there, to be discussed...

client/virtual_channel_util.go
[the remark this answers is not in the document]
sebastianst: In general, I'd agree that using interface map keys is a bad idea, e.g., if we were writing generic exported library code.

But here, we use the `state`s for a very specific user case, namely, for struct values, where it is totally fine. And since it would panic in the tests already if it wasn't fine, IMO it is ok to do it in this special use case. I mean, we don't even export the state watcher but only use it package-locally in very specific use cases.

client/virtual_channel_settlement.go
sebastianst: Ah so the update can still fail at this stage? Then yes. Is the virtual channel discarded from persistence after the parent channel update is successful?

client/virtual_channel_settlement.go
sebastianst: Why is is "Store"?

client/proposalmsgs_test.go
```diff
@@ -135,6 +141,23 @@ func TestSubChannelProposalReqProposalID(t *testing.T) {
...
 }
 
+func TestVirtualChannelProposalReqProposalID(t *testing.T) {
+	rng := pkgtest.Prng(t)
+
+	prop, err := clienttest.NewRandomVirtualChannelProposal(rng)
+	require.NoError(t, err)
+
+	testProp := *prop
+	assert.Equal(t, prop.ProposalID(), testProp.ProposalID())
```
ggwpez, line 151: Just to clarify:  
You use the embedded `BaseChannelProposal.ProposalID` here since `VirtualChannelProposal` does not have its own `ProposalID` function.  
Multiple different `VirtualChannelProposal`s *could* now have the same proposal ID, if the nonce is not unique.  

I am just thinking about if this is a problem or not, it seems weird but i think it is fine.  
Do you see a problem with it @sebastianst ?

pkg/io/test/serializertest.go
```diff
@@ -49,10 +50,9 @@ func genericDecodeEncodeTest(t *testing.T, serializers ...perunio.Serializer) {
 		r.Close()
 		if err != nil {
 			t.Errorf("failed to decode %dth element (%T): %+v", i, v, err)
-		} else if !reflect.DeepEqual(v, dest.Interface()) {
-			t.Errorf(
-				"encoding and decoding the %dth element (%T) resulted in different value: %v, %v",
-				i, v, reflect.ValueOf(v).Elem(), dest.Elem())
+		} else {
+			_v := dest.Interface()
+			assert.Equalf(t, v, _v, "comparing element %d", i)
```
ggwpez, line 55: Why did you change this?

wire/test/serializinglocalbus.go
```diff
@@ -0,0 +1,42 @@
+package test
```
ggwpez, line 1: License test missing

wire/test/serializinglocalbus.go
```diff
@@ -0,0 +1,42 @@
...
+		return
+	}
+
+	var _e wire.Envelope
+	err = _e.Decode(&buf)
+	if err != nil {
+		return
+	}
+
+	wire.TestMsg(b.t, e.Msg)
```
ggwpez, line 39: I think either En/decoding manually or calling this test should be enough.

client/virtual_channel.go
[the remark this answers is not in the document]
matthiasgeihs: after chat discussion, avoiding named returns as per convention

client/client_persistence_test.go
```diff
@@ -15,48 +15,22 @@
...
-	"perun.network/go-perun/pkg/test"
-	"perun.network/go-perun/wire"
 )
 
 func TestPersistencePetraRobert(t *testing.T) {
-	rng := test.Prng(t)
-	for i := 0; i < 2; i++ {
-		setups := NewSetupsPersistence(t, rng, []string{"Petra", "Robert"})
-		roles := [2]ctest.Executer{
+	runTwoPartyTest(t, func(rng *rand.Rand) (setups []ctest.RoleSetup, roles [2]ctest.Executer) {
```
ggwpez, line 26: Nice!

client/virtual_channel.go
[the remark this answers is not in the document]
matthiasgeihs: done

ggwpez (changes requested):
You could extract the BigInt En/Decoding and wire/test.LocalBus changes into another MR, if you want.

client/virtual_channel_settlement.go
matthiasgeihs: done

client/virtual_channel_settlement.go
matthiasgeihs: We also store the final state. I am currently unsure whether this is really needed. To be discussed [here](https://github.com/hyperledger-labs/go-perun/pull/123#discussion_r662263970).

client/virtual_channel_settlement.go
matthiasgeihs: Yes, because the accept handling comes afterwards.

Before we delete the channel, we need to persist the new parent channel states. This is only done during accept handling. Alternatively, we could do the accept handling here. Then we should be fine with deleting the channel state.

client/virtual_channel_util.go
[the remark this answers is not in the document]
matthiasgeihs: OK, created a corresponding commit and it seems to work.

client/proposal.go
```diff
@@ -390,7 +390,15 @@ func (c *Client) validVirtualChannelProposal(prop *VirtualChannelProposal, ourId
 		return errors.Errorf("expected %d index maps, got %d", numPeers, numIndexMaps)
 	}
 
-	virtualBals := transformBalances(prop.InitBals.Balances, parent.state().NumParts(), prop.IndexMaps[ourIdx])
+	// Check index map entries.
+	indexMap := prop.IndexMaps[ourIdx]
+	for i, p := range indexMap {
+		if int(p) >= numPeers {
```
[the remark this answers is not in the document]
matthiasgeihs: This here is just to prevent uninted usage. It will be changed in detail again in `matchFundingProposal`.

channel/test/randomizer.go
```diff
@@ -222,6 +222,33 @@ func NewRandomChannelID(rng *rand.Rand, opts ...RandomOpt) (id channel.ID) {
...
+		ids[i] = NewRandomChannelID(rng)
+	}
+	return
+}
+
+// NewRandomIndexMap generates a random index map.
+func NewRandomIndexMap(rng *rand.Rand, numParts int, numPartsParent int) (m []channel.Index) {
+	m = make([]channel.Index, numParts)
+	for i := range m {
+		m[i] = channel.Index(rng.Intn(numPartsParent))
```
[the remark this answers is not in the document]
matthiasgeihs: In fact, this is valid. In a multi-party virtual channel, the hub can take the role of several participants.

client/test/randomproposal.go
```diff
@@ -54,3 +54,17 @@ func NewRandomSubChannelProposal(rng *rand.Rand, opts ...client.ProposalOpts) (*
 		channeltest.NewRandomAllocation(rng, channeltest.WithNumParts(2)),
 		opts...)
 }
+
+// NewRandomVirtualChannelProposal creates a random virtual channel proposal with the
+// supplied options. Number of participants is fixed to 2.
+func NewRandomVirtualChannelProposal(rng *rand.Rand, opts ...client.ProposalOpts) (*client.VirtualChannelProposal, error) {
+	numParts := 2
```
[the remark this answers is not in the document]
matthiasgeihs: `NewRandomLedgerChannelProposal` and `NewRandomSubChannelProposal` do the same thing. I don't know why. Just was doing the same.

client/proposalmsgs_test.go
matthiasgeihs: I am doing that since it is unique do to the nonce. No need to encode the whole set of parameters.

pkg/io/test/serializertest.go
matthiasgeihs: `assert.Equal` gives information on where the objects differ. This can be very helpful.

wire/test/serializinglocalbus.go
matthiasgeihs: This covers two different cases. There could be an error in the implementation of `TestMsg`, so encoding and decoding is useful. On the other hand, encoding and decoding will not give more details about where the encoding may have failed.

client/client_persistence_test.go
matthiasgeihs: The linter was suddenly complaining about duplicate code...

client/virtual_channel_happy_test.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		t.Fatalf("creating virtual channel proposal: %v", err)
+	}
+
+	chAliceBobAlice, err := alice.ProposeChannel(ctx, vcp)
+	if err != nil {
+		t.Fatalf("opening channel between Alice and Bob: %v", err)
+	}
+
+	err = chAliceBobAlice.UpdateBy(ctx, func(s *channel.State) error {
```
sebastianst, line 171: Please do _a couple_ of channel updates (at least one) before settling so that the final state is >1. It would be good to test that the hub's channel machines handle skipped version numbers correctly, when the final state is set. I also hinted at this in yesterday's review.

sebastianst (changes requested):
I remember that in the other client tests, where we defined those `Role`s, we defined the proposal and update handlers on the `Role` classes and it contained channels on which we could get the resulting channel from a proposal accept/accept a channel update. This way, more logic was encapsulated into methods on the roles and there were no closures defined as handlers in the test's main code. It also made the roles more reusable.

client/subchannel.go
sebastianst: So the flag is not guaranteed to match that information?

client/virtual_channel_settlement.go
sebastianst: I'll reference this in an issue, so that we don't forget to plug the persistence leak.

client/test/randomproposal.go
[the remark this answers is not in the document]
sebastianst: As long as the `WithNumParts` option overwrites this, it'd be a sane default. Is that the case?

client/proposalmsgs_test.go
sebastianst: AFAIR, we don't use the proposal id in the protocol for security reasons, right? I mean, it's not a commitment to all proposal params like the channel id is a commitment to the channel params. We rather use it as a means to have a reference id in the proposal protocol. If that is the case, I think it's fine like this. If one participant changes some params, it shouldn't (hopefully) be a security concern because only what is eventually signed (first state) matters.

But maybe good to double check in how far we rely on the proposal id to be a commitment to the proposal parameters.

pkg/io/test/serializertest.go
sebastianst: I think it makes sense 👍🏻 We wrote this test very early on when we didn't use so much `assert`/`require`.

Note that it would now be nicer to `return` above after the `t.Error` and then move the `assert.Equalf` outside the `else` branch.

wire/test/serializinglocalbus.go
sebastianst: If you suspect `TestMsg` to have a bug in its implementation, it is better to write a proper test for `TestMsg` (can use the `pkg/test.Tester` for this) and then use it with full trust. This is what helpers like `TestMsg` are for, to avoid duplication.

client/proposalmsgs_test.go
matthiasgeihs: Yes, that is my understanding as well.

client/subchannel.go
matthiasgeihs: Not sure what you mean. There is no subchannel flag.

client/test/randomproposal.go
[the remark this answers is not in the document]
matthiasgeihs: Not sure what you mean by "As long as the WithNumParts option overwrites this".

pkg/io/test/serializertest.go
matthiasgeihs: This would change the functionality, right? Why is it nicer in your opinion?

wire/test/serializinglocalbus.go
matthiasgeihs: I don't really suspect it to have a bug, but who knows. Anyways, I will keep the wire.TestMsg and remove the encoding decoding above.

client/client_role_test.go
```diff
@@ -62,6 +65,26 @@ func NewSetups(rng *rand.Rand, names []string) []ctest.RoleSetup {
 	return setup
 }
 
+type Client struct {
+	*client.Client
+	ctest.RoleSetup
+}
```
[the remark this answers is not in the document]
matthiasgeihs: I know that we are using the roles in the other tests, but this has downsides.

1. It makes the tests hard to understand. I also had a lot of trouble in the past working with the roles because of this and debugging was difficult.
2. I don't believe the same roles work for the virtual channel tests, even though I didn't check in detail, and writing similar roles is just a lot of work and I don't see the benefits.

Besides that, I already simplified the tests in the following commits that are left. You can have a look at #83 .

client/test/randomproposal.go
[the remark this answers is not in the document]
matthiasgeihs: OK, I think I get it now. But this is not the case. It does as it is specifies. It fixes NumParts to 2. And as I said, it is the same for `NewRandomLedgerChannelProposal` and `NewRandomSubChannelProposal`.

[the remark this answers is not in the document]
matthiasgeihs: If this should be changed, I suggest to create an issue for that. This MR here is already blowing up.

client/subchannel.go
sebastianst: I thought _sub channel_ means, _not ledger channel_, and there's a flag for this in the params.

pkg/io/test/serializertest.go
sebastianst: Ah I didn't see that this is happening inside a loop. nvm.

client/virtual_channel_happy_test.go
```diff
@@ -0,0 +1,203 @@
...
+				chAliceBobBob <- _chAliceBobBob
+			default:
+				t.Fatalf("invalid channel proposal: %v", cp)
+			}
+		},
+		updateProposalHandler: func(s *channel.State, cu client.ChannelUpdate, ur *client.UpdateResponder) {
+			ur.Accept(ctx)
+		},
+	}
+	go bob.Client.Handle(proposalHandlerBob, proposalHandlerBob)
```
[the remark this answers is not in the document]
matthiasgeihs: using chan error as suggested as an alternative by you

client/virtual_channel_happy_test.go
```diff
@@ -0,0 +1,203 @@
...
+		ch := <-chAliceBobBob
+		errs <- ch.Settle(ctx, false)
+	}()
+
+	if err := <-errs; err != nil {
+		t.Fatalf("closing virtual channel: %v", err)
+	}
+	if err := <-errs; err != nil {
+		t.Fatalf("closing virtual channel: %v", err)
+	}
```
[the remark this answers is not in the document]
matthiasgeihs: doing something similar

matthiasgeihs:
Force push: revert big.Int serialization change.

client/virtual_channel_settlement.go
matthiasgeihs: Added a commit where the virtual channel is withdrawn and thereby removed from persistence.

matthiasgeihs:
Squashing and rebasing on request by @ggwpez .

ggwpez (approved):

sebastianst (approved):
LGTM

PR #124, opened 2021-06-25
Add functions to register new assets in ethereum funder implementation

manoranjith:
- As these methods provide convenience for the user (to register assets after funder is initialized) and are not required by the framework itself for the process of funding, they are not added to the `Funder` interface.

- Use the `RegisterAsset` method in place of `WithDepositor`, as the newly added method provides the same functionality without making a copy of the funder.

- Also, fixed a bug in `funder_test.go`. See the last before commit for details.

### DIFF
--- a/backend/ethereum/channel/funder.go
+++ b/backend/ethereum/channel/funder.go
@@ -76,18 +76,6 @@ func NewFunder(backend ContractBackend) *Funder {
 	}
 }
 
-// copy returns a copy of the funder.
-func (f *Funder) copy() (_f *Funder) {
-	_f = NewFunder(f.ContractBackend)
-	for k, v := range f.accounts {
-		_f.accounts[k] = v
-	}
-	for k, v := range f.depositors {
-		_f.depositors[k] = v
-	}
-	return
-}
-
 // RegisterAsset registers the depositor and account for the specified asset in
 // the funder.
 //
@@ -126,19 +114,6 @@ func (f *Funder) IsAssetRegistered(asset Asset) (Depositor, accounts.Account, bo
 		return f.depositors[asset], acc, true
 	}
 	return nil, accounts.Account{}, false
-}
-
-// WithDepositor creates a copy of the funder and assigns a depositor for the
-// specified asset. Transactions by the depositor will be done using the
-// specified account.
-func (f *Funder) WithDepositor(asset Asset, d Depositor, acc accounts.Account) (_f *Funder) {
-	f.mtx.Lock()
-	_f = f.copy()
-	f.mtx.Unlock()
-
-	_f.accounts[asset] = acc
-	_f.depositors[asset] = d
-	return
 }
 
 // Fund implements the channel.Funder interface. It funds all assets in
--- a/backend/ethereum/channel/funder.go
+++ b/backend/ethereum/channel/funder.go
@@ -17,6 +17,7 @@ package channel
 import (
 	"context"
 	"math/big"
+	"sync"
 
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/accounts/abi"
@@ -46,7 +47,14 @@ type assetHolder struct {
 }
 
 // Funder implements the channel.Funder interface for Ethereum.
+//
+// In addition to the `Fund` method required by the `Funder` interface, it also
+// provides additional functions for convenience.
+//
+// All the exported methods are thread-safe and can be invoked concurrently.
 type Funder struct {
+	mtx sync.RWMutex
+
 	ContractBackend
 	// accounts associates an Account to every AssetIndex.
 	accounts map[Asset]accounts.Account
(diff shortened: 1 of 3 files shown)


manoranjith:
@matthiasgeihs @ggwpez I noticed a piece of code, which was causing the tests to fail after adding a mutex to the funder.

Looking at similar tests, I thought it could be a bug and fixed it in commit 0092cb34c22f0f2b1f7077e3b141319511fa8da9

Can you review it in particular to see if it is a bug in the tests or intended behavior ?

backend/ethereum/channel/funder.go
```diff
@@ -59,6 +62,12 @@ type Funder struct {
 var _ channel.Funder = (*Funder)(nil)
 
 // NewFunder creates a new ethereum funder.
+//
+// In addition to the `Fund` method required by the `Funder` interface, it also
```
matthiasgeihs, line 66: I think this comment is not necessary and hard to maintain. Not necessary because the doc will also list the functions. And hard to maintain because if we change the function, we need to change this comment which is easy to miss.

backend/ethereum/channel/funder.go
```diff
@@ -59,6 +62,12 @@ type Funder struct {
...
+// providers two additional functions for convenience: RegisterAsset and
+// IsAssetRegistered.
```
matthiasgeihs, line 68: I think this comment is not necessary and hard to maintain. Not necessary because the doc will also list the functions. And hard to maintain because if we change the function, we need to change this comment which is easy to miss.

backend/ethereum/channel/funder.go
```diff
@@ -59,6 +62,12 @@ type Funder struct {
...
+//
+// All the exported methods are thread-safe and can be invoked concurrently.
```
matthiasgeihs, line 70: This should go on type funder, I would say.

backend/ethereum/channel/funder.go
```diff
@@ -68,26 +77,44 @@ func NewFunder(backend ContractBackend) *Funder {
...
+// specified account when funding. Hence, it is the responsibility of the
+// caller to ensure, the account has sufficient balance in the asset.
+//
+// It returns true if the asset was successfully registered, false if it was already
+// present.
+func (f *Funder) RegisterAsset(asset Asset, d Depositor, acc accounts.Account) bool {
+	f.mtx.Lock()
+	defer f.mtx.Unlock()
+
+	// Both the maps (f.accounts & f.assets) are always modified togethe such
```
matthiasgeihs, line 93: It would probably be a good idea to merge them, right? Maybe something for another PR.

backend/ethereum/channel/funder.go
```diff
@@ -349,6 +379,9 @@ func FundingIDs(channelID channel.ID, participants ...perunwallet.Address) [][32
 
 // NumTX returns how many Transactions are needed for the funding request.
 func (f *Funder) NumTX(req channel.FundingReq) (sum uint32, err error) {
+	f.mtx.Lock()
```
matthiasgeihs, line 382: You could use a read-write lock and allows concurrent reads. I think only `RegisterAsset` writes and should not be run concurrently. The other methods are probably fine running concurrently?

backend/ethereum/channel/funder_test.go
```diff
@@ -46,6 +46,44 @@ const (
...
+	ksWallet := wallettest.RandomWallet().(*keystore.Wallet)
+	cb := ethchannel.NewContractBackend(simBackend, keystore.NewTransactor(*ksWallet, types.NewEIP155Signer(big.NewInt(1337))))
+	funder := ethchannel.NewFunder(cb)
+	assets := make([]ethchannel.Asset, n)
+	depositors := make([]ethchannel.Depositor, n)
+	accs := make([]accounts.Account, n)
+	for i := 0; i < n; i++ {
+		assets[i] = *(wallettest.NewRandomAddress(rng).(*ethwallet.Address))
+		accs[i] = accounts.Account{Address: ethwallet.AsEthAddr(wallettest.NewRandomAddress(rng))}
+	}
```
matthiasgeihs, line 63: Can you move the setup code into a separate function so that the code of this method focuses on the actual test cases?

matthiasgeihs (commented):
Looks good overall, some minor comments / requests.

backend/ethereum/channel/funder.go
manoranjith: I agree with on the difficulty in maintaining this comment. Maybe we should change this to 
```
In addition to `Fund` method required by the `Funder` interface, it provides additional methods for convenience.
```

To indicate there are other methods ? Or can it be removed completely ?

manoranjith: If it is to be included, should these lines also be moved to the `Funder` type ?

backend/ethereum/channel/funder.go
manoranjith: Yes, we could also use a read write lock here. I will update the PR.

backend/ethereum/channel/funder.go
manoranjith: Yes, I think it is better to merge these two fields. I think it could be done as a separate PR.

backend/ethereum/channel/funder_test.go
manoranjith: Updated in this [force-push](https://github.com/hyperledger-labs/go-perun/compare/c5255fbe5f5453796953eb1c9e559a9565a7c2ea..4a12ab6d8062a3296b14d92a697dc705aec0b7e5).

backend/ethereum/channel/funder.go
manoranjith: Updated in this [force-push](https://github.com/hyperledger-labs/go-perun/compare/c5255fbe5f5453796953eb1c9e559a9565a7c2ea..4a12ab6d8062a3296b14d92a697dc705aec0b7e5).

backend/ethereum/channel/funder.go
manoranjith: Updated in this [force-push](https://github.com/hyperledger-labs/go-perun/compare/c5255fbe5f5453796953eb1c9e559a9565a7c2ea..4a12ab6d8062a3296b14d92a697dc705aec0b7e5).

backend/ethereum/channel/funder.go
manoranjith: Updated in this [force-push](https://github.com/hyperledger-labs/go-perun/compare/c5255fbe5f5453796953eb1c9e559a9565a7c2ea..4a12ab6d8062a3296b14d92a697dc705aec0b7e5).

matthiasgeihs (approved):
👌

PR #126, opened 2021-06-29
:bug: [client] Fix update error propagation

matthiasgeihs:
Fixes a bug where `updateGeneric` discarded a `PeerRejectedError` and replaced it with a different error.

### DIFF
--- a/client/update.go
+++ b/client/update.go
@@ -172,11 +172,7 @@ func (c *Channel) updateGeneric(
 	}
 	// if anything goes wrong from now on, we discard the update.
 	// TODO: this is insecure after we sent our signature.
-	defer func() {
-		if err != nil {
-			err = c.discardUpdate(ctx)
-		}
-	}()
+	defer func() { c.handleUpdateError(ctx, err) }()
 
 	sig, err := c.machine.Sig(ctx)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


client/update.go
```diff
@@ -221,11 +217,13 @@ func (c *Channel) updateGeneric(
...
-	if derr := c.machine.DiscardUpdate(ctx); derr != nil {
-		// discarding update should never fail
-		err = errors.WithMessagef(derr,
-			"progressing update failed: %v, then discarding update failed", err)
+func (c *Channel) handleUpdateError(ctx context.Context, updateErr error) (err error) {
+	if err != nil {
+		if derr := c.machine.DiscardUpdate(ctx); derr != nil {
+			// discarding update should never fail
+			err = errors.WithMessagef(derr,
+				"progressing update failed: %v, then discarding update failed", updateErr)
```
manoranjith, line 225:
```suggestion
                       if derr == nil {
                       return err
                       }
			return errors.WithMessagef(updateErr, "progressing update failed, then discarding update failed", derr)
```

In it's current form, when `derr` is `nil`, `errors.WithMessagef` will also return a nil. So `updateErr` will be missed even if it is not nil.

Separated into two returns because, when `derr` is nil, the phrase "then discarding update failed" will be confusing.

manoranjith (changes requested):

client/update.go
manoranjith: @matthiasgeihs

matthiasgeihs: derr cannot be nil there because `; derr != nil {`

manoranjith: Sorry, i missed that part. Then, it can be as it was before.

However, I think you may have to do `errors.Messagef(updateErr..` instead of `errors.Messagef(derr..`  Because if only the message of `updateErr` is included, I think it cannot be retrieved later using `errors.As`.

matthiasgeihs: OK. Changing the ordering seemed unclean. Instead I do according to your first suggestion, don't change the error, just log `derr`. Please have a look.

manoranjith:
LGTM. You will have to rebase to latest dev and then I can approve it.

manoranjith (approved):

PR #127, opened 2021-06-29
Draft channel redesign

matthiasgeihs:
This PR explores possibilities to streamline channel operation. Usage of the channel became more complicated since sub-channels and app channels were introduced.

One complication was caused by the fact that the register call was exposed and needed to be called manually before enforcing on-chain state progression.

This PR proposes to simplify the process of on-chain progression:

- Rename Channel.Progress to Channel.ForceUpdate: This conveys more clearly that the functionality is related to "Channel.Update".
- Implicitly register state on `ForceUpdate`: This simplifies the usage because the user does need to take care of manually registering before calling `ForceUpdate`.

The PR also includes parameter consolidation for `Adjudicator.Register` and `Withdraw`, similar to `Adjudicator.Progress`.

### DIFF
--- a/backend/ethereum/channel/adjudicator_test.go
+++ b/backend/ethereum/channel/adjudicator_test.go
@@ -88,13 +88,15 @@ func TestSubscribeRegistered(t *testing.T) {
 	defer sub.Close()
 	// Now test the register function
 	tx := testSignState(t, s.Accs, params, state)
-	req := channel.AdjudicatorReq{
-		Params: params,
-		Acc:    s.Accs[0],
-		Idx:    channel.Index(0),
-		Tx:     tx,
+	req := channel.RegisterReq{
+		AdjudicatorReq: channel.AdjudicatorReq{
+			Params: params,
+			Acc:    s.Accs[0],
+			Idx:    channel.Index(0),
+			Tx:     tx,
+		},
 	}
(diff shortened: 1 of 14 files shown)


manoranjith:
Hi @matthiasgeihs,

I think description does cover all the changes you have included in this PR. But i still misses one point (calling `registerDispute` within `Settle`. Below, I am summarizing the objective of this re-design for easier reference:

#### Objective of this re-design from usage perspective

Before this re-design:
1. `channel.Register` should be called before settling / progressing a channel.
2.  To settle a channel, call `ch.Settle` after a `ConcludedEvent` (in case of collaborative close) or a `RegisteredEvent` (in case of non-collaborative close) is received in the `AdjudicatorEventHandler`.

After this re-design:
1. Directly call `channel.Progress` or `channel.Settle` for progressing / settling a channel respectively.
Rename `channel.Progress` -> `channel.ForceUpdate` as the latter name conveys the meaning better.

#### My observations on the changes

1. I think the first part (to call `Register` within `channel.Progress` call is implemented.
2. But second part (to call 'Register` within `channel.Settle` needs to be implemented.
3. I also like the idea of implementing `ensureRegistered`, `waitForRegisterd` functions in the core abstraction rather than in the backend.

matthiasgeihs:
Parts of this draft was implemented in #143

PR #134, opened 2021-07-02
Fix tx nonce mismatch error: approach 1

### LINKED ISSUE #62: ERC20 Depositor nonce race
**Problem**: The `ERC20Depositor` sends two transactions (increase allowance and deposit) in the same block.  
This seems to be causing issues with how the nonce is determined. Sometimes its transactions revert.

**Possible solutions**:  
1. Manually increasing the nonce by one after the first TX
2. Wait for the first TX to be mined (which is actually the job of the `Funder`)
3. Poll the nonce multiple times until a new one is returned

I opt for `1.` as it seems the easiest and fastest.

manoranjith:
Fix is attempted using approach 1 suggested in #62.

### DIFF
--- a/backend/ethereum/channel/contractbackend.go
+++ b/backend/ethereum/channel/contractbackend.go
@@ -18,11 +18,13 @@ import (
 	"context"
 	stderrors "errors"
 	"math/big"
+	"sync"
 
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/pkg/errors"
 
@@ -60,7 +62,9 @@ type Transactor interface {
 // This is needed to send on-chain transaction to interact with the smart contracts.
 type ContractBackend struct {
 	ContractInterface
-	tr Transactor
+	tr                Transactor
+	nonceMtx          *sync.Mutex
+	expectedNextNonce map[common.Address]uint64
 }
 
 // NewContractBackend creates a new ContractBackend with the given parameters.
@@ -68,6 +72,8 @@ func NewContractBackend(cf ContractInterface, tr Transactor) ContractBackend {
 	return ContractBackend{
 		ContractInterface: cf,
 		tr:                tr,
+		expectedNextNonce: make(map[common.Address]uint64),
+		nonceMtx:          &sync.Mutex{},
 	}
 }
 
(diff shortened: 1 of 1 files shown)


matthiasgeihs:
LGTM, but there is a typo in your sign off.

manoranjith:
Fixed the signed-off message.

backend/ethereum/channel/contractbackend.go
```diff
@@ -130,6 +143,17 @@ func (c *ContractBackend) NewTransactor(ctx context.Context, gasLimit uint64,
 	auth.GasLimit = gasLimit
 	auth.Context = ctx
 
+	nonce, err := c.PendingNonceAt(ctx, acc.Address)
+	if err != nil {
+		return nil, errors.Wrap(err, "fetching nonce")
```
ggwpez, line 148: I think you have to check for `IsChainNotReachableError` here.

backend/ethereum/channel/contractbackend.go
```diff
@@ -122,6 +128,13 @@ func (c *ContractBackend) pastOffsetBlockNum(ctx context.Context) (uint64, error
 // set the value manually afterwards if it should be different from 0.
 func (c *ContractBackend) NewTransactor(ctx context.Context, gasLimit uint64,
 	acc accounts.Account) (*bind.TransactOpts, error) {
+	c.nonceMtx.Lock()
+	defer c.nonceMtx.Unlock()
+	expectedNextNonce, found := c.expectedNextNonce[acc.Address]
```
ggwpez, line 133: You use the `expectedNextNonce` value below. Does golang define how it is set if the map access fails?

ggwpez (changes requested):

backend/ethereum/channel/contractbackend.go
matthiasgeihs: Why?

backend/ethereum/channel/contractbackend.go
matthiasgeihs: [Default zero value](https://golang.org/ref/spec#The_zero_value)

backend/ethereum/channel/contractbackend.go
manoranjith: @matthiasgeihs  Because, we explicitly inform the user by returning a named error when there are problems connecting with the blockchain node. 

@ggwpez I have fixed this now.

matthiasgeihs: I see. It seems a bit unclean that we introduce such specific error handling logic in so many places. We could add wrappers for the handful of functions that need this handling and then consistently use the wrappers, right? It's ok here, but it's something worth discussing in general, I think.

ggwpez: Yes the initial approach calls the error wrapping on the lowest level, where we also could have done it on the highest…  
Its much more effort like this i think.

backend/ethereum/channel/contractbackend.go
ggwpez: Good to know.

ggwpez (approved):

PR #135, opened 2021-07-02
Fix tx nonce mismatch error: approach 3

### LINKED ISSUE #62: ERC20 Depositor nonce race

manoranjith:
Fix is attempted using approach 1 suggested in #62.

Polling until a new nonce is received, ensures, a transaction opts is created only after the previous transaction has reached the blockchain node.

As a results, tests for `NewTransactor` fail. In these tests, new transactor instances are created without sending the previous ones to blockchain node, causing the `PendingNonceAt` to return the same value and eventually timeout. This test is updated to use different random accounts for each `NewTransactor` call.

Resolves #62.

### DIFF
--- a/backend/ethereum/channel/contractbackend.go
+++ b/backend/ethereum/channel/contractbackend.go
@@ -18,11 +18,14 @@ import (
 	"context"
 	stderrors "errors"
 	"math/big"
+	"sync"
+	"time"
 
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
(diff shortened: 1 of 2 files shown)


manoranjith:
Since we have merged #134 to resolve #62, this can be closed ?
@matthiasgeihs @ggwpez

ggwpez:
> Since we have merged #134 to resolve #62, this can be closed ?  

AFAIK this both fixes the same issue, right? Then you can close it.

PR #144, opened 2021-07-20
143 settle implicit register

### LINKED ISSUE #143: Improve `Channel` UX: Register implicitly during `Settle`
**Location:** Package client, type `Channel`

**Problem:** Using type `Channel` became quite complicated since the introduction of sub-channels and app channels. Before, only function `Settle` was exposed in order to conclude and settle a channel. Now, additionally `Register` is exposed, and depending on the situation, it needs to be called before being able to settle a channel. Requiring the user to understand when to call `Register` and when not to makes the usage of `Channel` unnecessarily complicated.

**Suggestion:**
- Change `Settle` such that it automatically calls `register` when necessary.
- Change `ProgressBy` such that it automatically calls `register` when necessary.
  - Rename `ProgressBy ` to `ForceUpdate` to indicate its relation to `Update`.
- Unexport `Register`.
  - Rename `Register` to `registerDispute` to clarify that it is only relevant during dispute.

matthiasgeihs:
Closes #143

### DIFF
--- a/client/adjudicate.go
+++ b/client/adjudicate.go
@@ -134,13 +134,13 @@ func (c *Channel) registerDispute(ctx context.Context) error {
 	return nil
 }
 
-// ProgressBy progresses the channel state in the adjudicator backend.
+// ForceUpdate enforces a channel update on the adjudicator.
 //
 // Returns TxTimedoutError when the program times out waiting for a transaction
 // to be mined.
 // Returns ChainNotReachableError if the connection to the blockchain network
 // fails when sending a transaction to / reading from the blockchain.
-func (c *Channel) ProgressBy(ctx context.Context, update func(*channel.State)) error {
+func (c *Channel) ForceUpdate(ctx context.Context, update func(*channel.State)) error {
 	err := c.ensureRegistered(ctx)
 	if err != nil {
 		return err
--- a/client/adjudicate.go
+++ b/client/adjudicate.go
@@ -76,7 +76,7 @@ func (c *Channel) Watch(h AdjudicatorEventHandler) error {
 
 			// If local version greater than backend version, register local state.
 			if e.Version() < c.State().Version {
-				if err := c.register(ctx); err != nil {
+				if err := c.registerDispute(ctx); err != nil {
 					return errors.WithMessage(err, "registering")
 				}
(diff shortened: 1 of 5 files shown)


client/adjudicate.go
```diff
@@ -396,3 +409,55 @@ func (c *Channel) subChannelStateMap() (states channel.StateMap, err error) {
 	})
 	return
 }
+
+// ensureRegistered ensures that the channel is registered.
+func (c *Channel) ensureRegistered(ctx context.Context) (err error) {
+	phase := c.Phase()
+	if phase == channel.Registered || phase == channel.Progressed {
+		return
```
manoranjith, line 417:
```suggestion
		return nil
```

This is a minor suggestion.

Generally, I feel it is easier to follow when you use explicit returns instead of named returns.

In the latter case, I will have to go up to the function definition to see the return variables and then figure where these variables have been set. 
But when using explicit returns, most information is contained within that one line.

client/adjudicate.go
```diff
@@ -396,3 +409,55 @@ func (c *Channel) subChannelStateMap() (states channel.StateMap, err error) {
...
+	select {
+	case e := <-registeredEvents:
+		err = e.Timeout().Wait(ctx)
+		if err != nil {
+			return err
+		}
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+	return
```
manoranjith, line 444: Same as above.

manoranjith (commented):
LGTM. Have only one minor suggestion..

sebastianst (dismissed):
LGTM. As discussed before (and mentioned by @manoranjith), explicit returns are better to read than named returns, and more conventional.

matthiasgeihs:
> LGTM. As discussed before (and mentioned by @manoranjith), explicit returns are better to read than named returns, and more conventional.

Parts of the code I already wrote some time ago, and I missed updating that part. Changing that now.

matthiasgeihs:
Rebase

matthiasgeihs:
Rebased and comments addressed.

manoranjith (approved):
LGTM.

sebastianst (approved):

Issue #102, opened 2021-06-11
Extended Event filtering


ggwpez:
**Location**: `[backend/eth/subscription]`  

**Problem**: The event filtering of *go-ethereum* allows to pass a query.  
A query is a set of sets of arguments, where at least on set of arguments must match. Missing parameters are treated as wildcard.   

The problem is that eg. in the `Adjudicator` we want to filter for `ChannelUpdate` events with a specific `ChannelID` and `Phase`.  
Since these are the first and third arguments, it does not seem possible to directly filter for them. It would work if they were the first and second arguments though.

**Solution**: One possibility would be to replacing the `Filter [][]interface{}` field with an easier to use `Predicate func(Event) bool` function pointer.

PR #148, opened 2021-07-22
:bug: [backend/eth] conclude after progression into final

### LINKED ISSUE #147: ensureConcluded fails when channel has been progressed to final
**Location:** backend/eth/conclude, contracts/Adjudicator.sol

**Problem:** When a channel has been force-update on-chain into a final state, the participants will not have fully signed states. Now, when calling `ensureConcluded`, it will call `concludeFinal`, which expects a fully-signed state and thus fails.

**Idea:** If a channel has been force-updated into final, the client should call `conclude` instead of `concludeFinal`, because `conclude` does not require signatures. The client must be able to detect whether a channel is in a final state on-chain. Usually, we indicate that by setting `phase = CONCLUDED`. Hence, I suggest that we change the contract such that if a state is force-updated to final, we automatically set `phase = CONCLUDED`.

Update: We can also apply a quick fix that doesn't require a change in the contracts. We can check whether the current on-chain phase is `FORCE_EXECUTION`. If this is the case, we know that the state resulted from force execution, which means we do not have the all signatures and need to call `conclude` instead of `concludeFinal`.

matthiasgeihs:
Closes #147

### DIFF
--- a/backend/ethereum/channel/conclude.go
+++ b/backend/ethereum/channel/conclude.go
@@ -54,8 +54,8 @@ func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.Adjudicat
 	subErr := make(chan error, 1)
 	waitCtx, cancel := context.WithCancel(ctx)
 	go func() {
+		defer cancel()
 		subErr <- sub.Read(ctx, events)
-		cancel()
 	}()
 
 	// In final Register calls, as the non-initiator, we optimistically wait for
@@ -71,17 +71,12 @@ func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.Adjudicat
 	}
 
 	// No conclude event found in the past, send transaction.
-	if req.Tx.IsFinal {
-		err = errors.WithMessage(a.callConcludeFinal(ctx, req), "calling concludeFinal")
-	} else {
-		err = errors.WithMessage(a.callConclude(ctx, req, subStates), "calling conclude")
-	}
-	if IsErrTxFailed(err) {
-		a.log.Warn("Calling conclude(Final) failed, waiting for event anyways...")
-	} else if err != nil {
-		return err
+	err = a.conclude(ctx, req, subStates)
+	if err != nil {
+		return errors.WithMessage(err, "concluding")
 	}
 
+	// Wait for concluded event.
 	for {
 		select {
 		case _e := <-events:
@@ -100,6 +95,25 @@ func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.Adjudicat
 	}
 }
 
+func (a *Adjudicator) conclude(ctx context.Context, req channel.AdjudicatorReq, subStates channel.StateMap) error {
+	// If the on-chain state resulted from forced execution, we do not have a fully-signed state and cannot call concludeFinal.
+	forceExecuted, err := a.isForceExecuted(ctx, req.Params.ID())
+	if err != nil {
+		return errors.WithMessage(err, "checking force execution")
+	}
+	if req.Tx.IsFinal && !forceExecuted {
+		err = errors.WithMessage(a.callConcludeFinal(ctx, req), "calling concludeFinal")
+	} else {
+		err = errors.WithMessage(a.callConclude(ctx, req, subStates), "calling conclude")
+	}
+	if IsErrTxFailed(err) {
+		a.log.WithError(err).Warn("Calling conclude(Final) failed, waiting for event anyways...")
+	} else if err != nil {
+		return err
+	}
+	return nil
+}
+
 // isConcluded returns whether a channel is already concluded.
 func (a *Adjudicator) isConcluded(ctx context.Context, sub *subscription.EventSub) (bool, error) {
 	events := make(chan *subscription.Event, 10)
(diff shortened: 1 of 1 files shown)


backend/ethereum/channel/conclude.go
```diff
@@ -119,6 +133,30 @@ func (a *Adjudicator) isConcluded(ctx context.Context, sub *subscription.EventSu
...
+	defer sub.Close()
+	events := make(chan *subscription.Event, 10)
+	subErr := make(chan error, 1)
+	// Write the events into events.
+	go func() {
+		defer close(events)
+		subErr <- sub.ReadPast(ctx, events)
+	}()
+	// Read all events and check for force execution.
+	for _e := range events {
```
ggwpez, line 151: Looking at this manual filter logic really make me want implement #102 :laughing:

backend/ethereum/channel/conclude.go
```diff
@@ -119,6 +133,30 @@ func (a *Adjudicator) isConcluded(ctx context.Context, sub *subscription.EventSu
...
+	if err != nil {
+		return false, errors.WithMessage(err, "subscribing")
+	}
+	defer sub.Close()
+	events := make(chan *subscription.Event, 10)
+	subErr := make(chan error, 1)
+	// Write the events into events.
+	go func() {
+		defer close(events)
+		subErr <- sub.ReadPast(ctx, events)
```
ggwpez, line 148: I think in theory this would need a sub-ctx as context, to ensure that it gets cancelled on function exit.  
In practice there will probably not be enough events anyway.

backend/ethereum/channel/conclude.go
```diff
@@ -119,6 +133,30 @@ func (a *Adjudicator) isConcluded(ctx context.Context, sub *subscription.EventSu
...
+	}()
+	// Read all events and check for force execution.
+	for _e := range events {
+		e := _e.Data.(*adjudicator.AdjudicatorChannelUpdate)
+		if e.Phase == phaseForceExec {
```
ggwpez, line 153: Does it matter that you do not check all events? Can something after the force-exec happen?

backend/ethereum/channel/conclude.go
```diff
@@ -100,6 +95,25 @@ func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.Adjudicat
...
+		return errors.WithMessage(err, "checking force execution")
+	}
+	if req.Tx.IsFinal && !forceExecuted {
+		err = errors.WithMessage(a.callConcludeFinal(ctx, req), "calling concludeFinal")
+	} else {
+		err = errors.WithMessage(a.callConclude(ctx, req, subStates), "calling conclude")
+	}
+	if IsErrTxFailed(err) {
+		a.log.Warn("Calling conclude(Final) failed, waiting for event anyways...")
+	} else if err != nil {
```
ggwpez, line 111: I dont think you need this _else if_.

backend/ethereum/channel/conclude.go
```diff
@@ -100,6 +95,25 @@ func (a *Adjudicator) ensureConcluded(ctx context.Context, req channel.Adjudicat
...
+	if err != nil {
+		return errors.WithMessage(err, "checking force execution")
+	}
+	if req.Tx.IsFinal && !forceExecuted {
+		err = errors.WithMessage(a.callConcludeFinal(ctx, req), "calling concludeFinal")
+	} else {
+		err = errors.WithMessage(a.callConclude(ctx, req, subStates), "calling conclude")
+	}
+	if IsErrTxFailed(err) {
+		a.log.Warn("Calling conclude(Final) failed, waiting for event anyways...")
```
ggwpez, line 110: Please log the error, since it is also not returned.

ggwpez (changes requested):
Only one change request, others are comments.

backend/ethereum/channel/conclude.go
matthiasgeihs: This was not changed, just moved, but I will consider logging the error.

backend/ethereum/channel/conclude.go
matthiasgeihs: we do, because we don't want to fail, if it's an `ErrTxFailed`.

backend/ethereum/channel/conclude.go
matthiasgeihs: ok, let's use a sub-context then.

backend/ethereum/channel/conclude.go
matthiasgeihs: feel free :)

backend/ethereum/channel/conclude.go
matthiasgeihs: Currently, the only phase after FORCE_EXEC is CONCLUDED. Anyhow, it's a good observation. I will make sure that I check the phase of the last event, because this was the intention.

matthiasgeihs:
@ggwpez  Addressed your comments, please have another look.

ggwpez (commented):
Looks good! I assume you still squash.

matthiasgeihs:
Rebased, please check and merge.

ggwpez (approved):

PR #150, opened 2021-07-23
Event filter predicate in EventSub

### LINKED ISSUE #102: Extended Event filtering

ggwpez:
Closes #102


ggwpez:
Not important right now.

PR #162, opened 2021-08-20
Make EventSub a Closer

### LINKED ISSUE #161: Make EventSub a Closer
To be honest, it might have been better to move the introduction of `Closer` to a separate PR altogether.

_Originally posted by @matthiasgeihs in https://github.com/hyperledger-labs/go-perun/pull/116#discussion_r692265641_

ggwpez:
Using a `[pkg/sync] Closer` makes the `EventSub.Close()` call idempotent and easier to use.

Closes #161

### DIFF
--- a/backend/ethereum/subscription/eventsub.go
+++ b/backend/ethereum/subscription/eventsub.go
@@ -24,6 +24,8 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/event"
 	cherrors "perun.network/go-perun/backend/ethereum/channel/errors"
+	"perun.network/go-perun/log"
+	pkgsync "perun.network/go-perun/pkg/sync"
 )
 
 type (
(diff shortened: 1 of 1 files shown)


ggwpez:
While checking the tests of another MR, i saw this log message which is introduced by this MR:  
`msg="could not close EventSub" error="<nil>"`  
It seems that `IsAlreadyClosedError` does also return true for certain `nil` errors, this seems like a bug.

backend/ethereum/subscription/eventsub.go
```diff
@@ -199,25 +205,25 @@ func (s *EventSub) readFuture(ctx context.Context, sink chan<- *Event) error {
...
+		case <-s.closer.Closed():
 			return nil
 		}
 	}
 }
 
 // Close closes the sub and frees associated resources.
-// Should be called exactly once and panics otherwise.
-// Must not be called if the construction function returned with an error.
+// Can be called more than once, but should be called at least once.
```
matthiasgeihs, line 223: I think the second part is clear.
```suggestion
// Can be called more than once.
```

matthiasgeihs (approved):
Just a minor comment, which is not mandatory.

matthiasgeihs (approved):

PR #164, opened 2021-08-24
:construction_worker: [CI/github] Use caching

### LINKED ISSUE #160: Use caching in go CI
**Observation**:  The go CI tests take a long time.  

**Assumed cause**: The tests always pull all dependencies (1), builds the code every time (2) and reruns all tests (3).

**Possible solution**: Using caching between the pipeline runs should fix all issues and speed up the CI considerably.

matthiasgeihs:
Closes #160

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,15 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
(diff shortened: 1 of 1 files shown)


ggwpez (commented):
It seems to cache the dependencies, but not the test run.  
Just like when running locally without a change it just prints `ok (cached)` and exits.  
This would speed up the tests by ~3 min when no change occurs.  
What do you think @matthiasgeihs ?

matthiasgeihs:
> It seems to cache the dependencies, but not the test run.
> Just like when running locally without a change it just prints `ok (cached)` and exits.
> This would speed up the tests by ~3 min when no change occurs.
> What do you think @matthiasgeihs ?

Left that out at first, because it wasn't working correctly. Included it now with modification to the race test cache key.

ggwpez (approved):
Total duration: 31s  
looks good!

PR #175, opened 2021-09-01
 Modify Adjudicator.Subscribe to take channel.ID instead of channel.Params

### LINKED ISSUE #173: Adjudicator.Subscribe takes `channel.Params` while only `channel.ID` is needed
#### Situation: 
[`Adjudicator.Subscribe`](https://github.com/hyperledger-labs/go-perun/blob/06c5c9577b85ee0e3b7147cd19ff68ec54e5e569/channel/adjudicator.go#L63) API  currently takes `channel.Params` as an argument. However, looking at the implementation of [`Subscribe`](https://github.com/hyperledger-labs/go-perun/blob/06c5c9577b85ee0e3b7147cd19ff68ec54e5e569/backend/ethereum/channel/subscription.go#L36) for ethereum, I noticed only the `ID` field in `params` is used. 

#### Problem

I noticed this when implementing the watcher for IoT use cases (#172). In the design of Watcher described in proposal#004, the client should register the channel by sending the `channel.ID`. Watcher would then subscribe for adjudicator events for this channel using this channel ID.

Because, the `Subscribe` API requires `channel.Params`, it will be required to pass `channel.Params` all over, while it needs only `channel.ID`.

#### Possible solution

Modify the `Adjudicator.Subscribe` API to take `channel.ID` instead of `channel.Params`

manoranjith:
- Previously, Adjudicator.Subscribe was taking channel.Params, while it
  was using only the ID contained in the params.

- So, modified the API to take only the required parameter.

Resolves #173.

### DIFF
--- a/backend/ethereum/channel/adjudicator_test.go
+++ b/backend/ethereum/channel/adjudicator_test.go
@@ -73,7 +73,7 @@ func TestSubscribeRegistered(t *testing.T) {
 	// Set up subscription
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	registered, err := s.Adjs[0].Subscribe(ctx, params)
+	registered, err := s.Adjs[0].Subscribe(ctx, params.ID())
 	require.NoError(t, err, "Subscribing to valid params should not error")
 	// we need to properly fund the channel
 	txCtx, txCancel := context.WithTimeout(context.Background(), defaultTxTimeout)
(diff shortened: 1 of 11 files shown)


ggwpez (commented):
I think this is reasonable. It also looks like it would not cause issues with the two other backends that we are currently implementing.  
Do you agree @matthiasgeihs ?

matthiasgeihs (approved):

Issue #179, opened 2021-09-06
☂️ Beauty fixes


ggwpez:
List of oddities that we noticed in *go-perun* will be collected here for triage:
- [ ] 1. `channel.Nonce` is modelled as a `*big.Int` but describes a `[]byte` which can cause issues with negative number encoding.
- [ ] 2. `NewFundingReq` returns a `*FundingReq` (per convention) but `Funder.Fund` accepts a `FundingReq`.
- [ ] 3. [eth/Backend.VerifySignature](https://github.com/hyperledger-labs/go-perun/blob/5340aa3f6d9dc59739ad77cd362e0962de435024/backend/ethereum/wallet/backend.go#L69) contains a superfluous `if`
- [ ] 4. `wallet.Address` needs only to be an `Encoder`, not `Serializer`. Decoding is handled by the backend anyway.

Please extend this list.  
:point_right: Related issues will be moved to Milestone *Beautification*.

matthiasgeihs:
- `Params.ID` should be calculated on the fly to avoid inconsistencies. Members of `Params` are exported, but should be read only. 
- `AdjudicatorReq.Acc` and `AdjudicatorReq.Idx` are not needed for `Adjudicator.Register`. We could create `type RegisterReq` specifically for registering.

manoranjith:
Do you think we should rename `AdjudicatgorSubscription` to `AdjudicatorSub` ?
This would make it more concise and also in line with the name of other type `AdjudicatorReq`.

manoranjith:
> 
> 
>     * `Params.ID` should be calculated on the fly to avoid inconsistencies. Members of `Params` are exported, but should be read only.
> 
>     * `AdjudicatorReq.Acc` and `AdjudicatorReq.Idx` are not needed for `Adjudicator.Register`. We could create `type RegisterReq` specifically for registering.
> 
>     * `channel.MakeStateMap` could accept parameter `states ...*State`.
> 
>     * `Adjudicator.Withdraw` could take the beneficiary as a parameter. (Currently, the beneficiary is a property of the adjudicator.)

+1 Especially on `ReqisterReq` and `beneficiary` parameter for `Withdraw`.

In the watcher task, I currently fill dummy values for `Acc` and `Idx`, because I figured out this fields are not used.
Doing otherwise meant, I will have to provide this data to the watcher via a the `StartWatching` API. Which was unnecessary.

ggwpez:
[...]

PR #182, opened 2021-09-09
174 Implement local watcher

### LINKED ISSUE #174: Implement a local watcher
### Context

This is step 1 of 3 in refactoring the Watcher for IoT use cases (#172). Covers defining the interfaces for watcher and implementing a local watcher.

### Details

The interfaces and functionality of the watching component has been described in proposal [#004](https://github.com/hyperledger-labs/perun-proposals/blob/main/design/004-IoT-Adoption-Watcher.md)

The interface definitions can be placed in a new package `go-perun/watcher` and the local watcher implementation can be placed in `go-perun/watcher/local` package.

manoranjith:
### Overview

- This PR implements the server side component local watcher as per the design described in [proposal#4](https://github.com/hyperledger-labs/perun-proposals/blob/main/design/004-IoT-Adoption-Watcher.md). 
- This component should be initialized by the user of the framework and pass it as an argument to the `client.New` API.

### Implementation details

- The interface definitions are located in a package: `go-perun/watcher`.
- The implementation of local watcher is located in a `go-perun/watcher/local`.

#### Data structures

1. **Watcher**: 
    - This represents the watcher instance, initializes using the `NewWatcher` function.
    - It contains 
        - `channel.RegistererSubscriber` instance for interacting with the blockchain.
        - `registry` that holds a map of all the channels currently registered with the Watcher.
    - Both the above data structures are safe for concurrent use.

2. **ch**:
    - Represents the data corresponding to a channel.
    - It contains
        - `id` of the channel.
        - `params` of the channel. Required while registering dispute for the channel. This will be passed when registering the channel with the `Watcher` via `StartWatching` function.
        - `parent` of this channel. For ledger channels, this is set to `channel.Zero`.

        - `subChsAccess`, (relevant only for ledger channel) a mutex used for concurrency safe access of a family of channels. When a dispute registration is required for a channel, we traverse back to the parent and lock this mutex. It is unlocked only after the dispute registration is completed. This also ensures, we lock only the relevant channels and not the entire registry when registering disputes.

        - `registeredVersion`: when a dispute is registered for a ledger, sub or virtual channel, this field is updated. This is used to ensure, we do not register the same version more than once.
        - `requestLatestTx` and `latestTx` are a set of channels used for retrieving the latest state of this channel without use of mutex locks.

3. **registry**: Used to store and retrieve channels registered with the watcher. It is protected by a mutex.

4. **StatesPubSub** and **AdjudicatorPubSub** implements the functionalities exactly as described in the proposal.
        
#### Methods

1. `NewWatcher` initializes the watcher.
2. `Watcher.StartWatchingForLedgerChannel` and Watcher.StartWatchingForNonLedgerChannel` APIs are used to register a channel with the watcher. Though the implementations of both the APIs are same, two APIs are provided because
    - LedgerChannel does not require parent while,
    - Sub-Channel requires a parent.
    - Regarding the names, I also think we need a better name, used this for time being. 
3. `AdjudicatorSub.Next` can be used by the client for receiving adjudicator events from the watcher.
4. `StatesPub.Publish` can be used by the client to send newer transactions to the watcher.
5. `Watcher.StopWatching` and `Watcher.Shutdown` are yet to implemented.

#### Tests

It was difficult for me to figure out how to add tests using the simulated backend. Particularly, I couldn't figure out how to make the watcher send events to the watcher.

However, I implemented some tests based on generated mocks. This covers both happy path and most of the corner cases. I used these to guide my development. I have added them in the PR (in separate commits). If it is possible, we could replace them with tests using simulated backend or remove them (if they don't fit the testing style used in the project).

#### Resolves #174


matthiasgeihs (commented):
Just a few high level comments. Could you give me and @ggwpez a code intro in our next meeting?

watcher/watcher.go
```diff
@@ -0,0 +1,73 @@
...
+	// to the watcher.
+	//
+	// This is initialized when client starts watching for a given channel.
+	// Client can send each state by calling the publish method.
+	//
+	// The publisher will be closed when client requests the watcher to stop
+	// watching or when there is an error. After the client requests the
+	// watcher to stop watching, Publish method must not be called.
+	StatesPub interface {
+		Publish(channel.Transaction) error
```
ggwpez, line 56: How about a context here?

watcher/watcher.go
```diff
@@ -0,0 +1,73 @@
...
+	// from the watcher to the client.
+	//
+	// This is initialized when client starts watching for a given channel.
+	// Client can receive the event via the channel returned by Next method.
+	//
+	// This channel will be closed when client requests the watcher to stop
+	// watching or when there is an error. After the channel is closed, error
+	// message can be read using the Err method.
+	AdjudicatorSub interface {
+		Next() <-chan channel.AdjudicatorEvent
```
ggwpez: Did you have problems with implementing this interface?  
I often think about on how to best do subscriptions in golang.

ggwpez: Please add a comment that this function always returns the same channel.

watcher/watcher.go
ggwpez: Could you move some of the documentation from the interface to the functions?  
When a user uses the interface he will most likely look at the functions doc.

watcher/watcher.go
```diff
@@ -0,0 +1,73 @@
...
+	// latest state on the blockchain.
+	//
+	// Once a channel is closed, client can de-register the channel from the
+	// watcher by calling the StopWatching API. This will close the
+	// AdjudicatorSub and StatesSub. Client must not publish states after
+	// this.
+	Watcher interface {
+		StartWatchingLedgerChannel(context.Context, channel.SignedState) (StatesPub, AdjudicatorSub, error)
+		StartWatchingSubChannel(context.Context, channel.ID, channel.SignedState) (StatesPub, AdjudicatorSub, error)
+		StopWatching(channel.ID) error
```
ggwpez, line 43: Does this function contain network logic? I think a context would be future proof.

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,80 @@
...
+	if s.closed.IsSet() {
+		return s.err
+	}
+	s.err = errors.WithStack(ErrAlreadyClosed)
+	close(s.pipe)
+	s.closed.Set()
+	return nil
+}
+
+func (s *statesPubSub) statesStream() <-chan channel.Transaction {
```
ggwpez, line 74: We talked about back-porting the name of this to an earlier commit.

watcher/local/registry.go
```diff
@@ -0,0 +1,73 @@
...
+// retrieveUnsafe retrieves the channel from registry, expecting the caller to
+// have locked the registry.
+func (r *registry) retrieveUnsafe(id channel.ID) (*ch, bool) {
+	ch, ok := r.chs[id]
+	return ch, ok
+}
+
+// retrieve retrieves the channel from registry.
+func (r *registry) retrieve(id channel.ID) (*ch, bool) {
+	r.mtx.Lock()
```
ggwpez, line 61: Use the lock function here?

watcher/local/registry.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func newRegistry() *registry {
+	return &registry{
+		chs: make(map[channel.ID]*ch),
+	}
+}
+
+// lock locks the registry.
+func (r *registry) lock() {
```
ggwpez, line 37: I dont think this function is necessary.

watcher/local/watcher.go
```diff
@@ -15,12 +15,92 @@
...
+	statesPubSub := newStatesPubSub()
+	eventsToClientPubSub := newAdjudicatorEventsPubSub()
+	ch := newCh(id, parent, signedState.Params)
+
+	w.registry.addUnsafe(ch)
+
+	return statesPubSub, eventsToClientPubSub, nil
+}
+
+func newCh(id, parent channel.ID, params *channel.Params) *ch {
```
ggwpez: Can you move this to the file that defines `ch`?

watcher/local/watcher.go
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+package local
+
+import (
+	"perun.network/go-perun/channel"
+)
+
+type (
+	ch struct {
```
ggwpez, line 58: You normally name the types self explaining, i think `channel` would make it more understandable.

watcher/local/watcher.go
```diff
@@ -15,12 +15,92 @@
...
+	Config struct {
+		RegisterSubscriber channel.RegisterSubscriber
+	}
+
 	ch struct {
 		id     channel.ID
+		params *channel.Params
 		parent channel.ID
+
+		subChsAccess sync.Mutex
```
ggwpez, line 74: Can you please document when this mutex needs to he locked? So to say, what does it guard?

watcher/local/watcher_test.go
```diff
@@ -0,0 +1,100 @@
...
+	ethChannel "perun.network/go-perun/backend/ethereum/channel"
+	_ "perun.network/go-perun/backend/ethereum/channel/test" // For initilizing channeltest
+	"perun.network/go-perun/channel"
+	channeltest "perun.network/go-perun/channel/test"
+	"perun.network/go-perun/pkg/test"
+	"perun.network/go-perun/watcher/internal/mocks"
+	"perun.network/go-perun/watcher/local"
+)
+
+// func init() {
```
ggwpez, line 40: Move to `init_test.go` and change log level to `warn` or delete it if you dont need it.

watcher/local/watcher.go
```diff
@@ -104,3 +107,16 @@ func newCh(id, parent channel.ID, params *channel.Params) *ch {
 		parent: parent,
 	}
 }
+
+func (w *Watcher) handleEventsFromChain(eventsFromChainSub channel.AdjudicatorSubscription, eventsToClientPubSub adjudicatorPub,
+	thisCh *ch) {
+	for e := eventsFromChainSub.Next(); e != nil; e = eventsFromChainSub.Next() {
```
ggwpez, line 216: `eventsFromChainSub.Err()` is unused.  
We could also change the `Next()` function to return `(event, error)`  so that correct usage is enforced.

watcher/local/watcher.go
```diff
@@ -25,6 +26,12 @@ import (
 	"perun.network/go-perun/watcher"
 )
 
+const (
+	// Duration for which the watcher will wait for latest transactions
+	// when an adjudicator event is received.
+	statesFromClientWaitTime = 1 * time.Millisecond
```
ggwpez, line 33: I think this should be configurable, what do you think @matthiasgeihs ?  
It would be already good for debugging when using it in deployment.

watcher/local/watcher.go
```diff
@@ -95,28 +106,156 @@ func (w *Watcher) startWatching(ctx context.Context, parent channel.ID, signedSt
...
+			Params: subCh.params,
+			State:  subChTx.State,
+			Sigs:   subChTx.Sigs,
+		}
+	}
+	return subStates
+}
+
+func makeAdjudicatorReq(params *channel.Params, tx channel.Transaction) channel.AdjudicatorReq {
+	return channel.AdjudicatorReq{
```
ggwpez, line 299: Just a comment: I added an entry for adding `NewAdjudicatorReq` in https://github.com/hyperledger-labs/go-perun/issues/179.

watcher/local/watcher.go
```diff
@@ -33,6 +34,16 @@ const (
 )
 
 type (
+	// closer is the interface that wraps the close method.
+	closer interface {
+		close() error
+	}
+
+	// Closer is the interface that wraps the Close method.
+	Closer interface {
```
ggwpez, line 43: Im not really happy with the ad-hoc redefenition of a public `Closer` interface here.  
Please consider using `[pkg/sync] Closer` which also gives so some guarantees as well.

ggwpez (commented):
Comments from the meeting.

watcher/watcher.go
```diff
@@ -60,13 +60,14 @@ type (
...
-	// Client can receive the event via the channel returned by Next method.
+	// Client can receive the event via the channel returned by EventStream
+	// method.
 	//
 	// This channel will be closed when client requests the watcher to stop
 	// watching or when there is an error. After the channel is closed, error
 	// message can be read using the Err method.
 	AdjudicatorSub interface {
-		Next() <-chan channel.AdjudicatorEvent
+		EventStream() <-chan channel.AdjudicatorEvent
```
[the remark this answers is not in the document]
manoranjith: Updated documentation that `Multiple calls to the EventStream method returns the same channel instance.` in PR #212.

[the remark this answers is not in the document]
manoranjith: I think multiplexing is not necessary in our use case, because the only consumer of the events is the watch function.

If at a later point in time, we have the need for consuming the events in multiple places, then we could implement multiplexing.

watcher/local/watcher.go
```diff
@@ -104,3 +107,16 @@ func newCh(id, parent channel.ID, params *channel.Params) *ch {
 		parent: parent,
 	}
 }
+
+func (w *Watcher) handleEventsFromChain(eventsFromChainSub channel.AdjudicatorSubscription, eventsToClientPubSub adjudicatorPub,
+	thisCh *ch) {
+	for e := eventsFromChainSub.Next(); e != nil; e = eventsFromChainSub.Next() {
+		switch e.(type) {
+		case *channel.RegisteredEvent:
```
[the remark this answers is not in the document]
manoranjith: Updated in #215.

watcher/watcher.go
manoranjith: Have added the comment.

manoranjith: Yes, one aspect specific to go-channels is that
1. It is not possible to check if a channel is open, without reading from the channel.
2. Writing to a closed channel will cause panic.

So, the common approach is that, a channel must be closed only by the writer and never by the reader. Also, if there are multiple writers, then they must co-ordinate with each other while closing to ensure no writer writes after the channel has been closed.

Reader can check if a channel is closed by using `data, ok := <-ch` instead of `data := <-ch`. Because, when channel is closed, data will always be zero value and ok will be false.

manoranjith: [Here](https://go101.org/article/channel-closing.html) is a bit long, but very comprehensive article relating to closing of channels. @ggwpez

watcher/watcher.go
manoranjith: This function will contain network logic (subscribing to blockchain events).

watcher/local/statespubsub.go
manoranjith: Yes, I have done that. Now the term `EventStream` is used in `watcher/watcher.go` in the initial commit for that file. See #212.

watcher/local/registry.go
manoranjith: I did not get your point ? @ggwpez

watcher/local/watcher.go
manoranjith: Currently, this delay is used only to account for receiving events that are already published, but not read by the watcher. That is also the reason, why this timeout is very small.

However, as we had discussed in the call, in future we could introduce the concept of `Sync interval` and once we do that, we will have to make this as a configurable parameter of the watcher.

watcher/local/watcher.go
manoranjith: I think, in that case, `NewAdjudicatorReq` should be implemented as a function that takes `params` and `Transaction` as parameter and constructs an `AdjudicatorReq` rather than being a method over channel type.

Because,
1. In case of client, it might never need this method.
2. In case of watcher, the watcher will not have access to channel instance, rather only to `params` and `Transaction`.

channel/adjudicator.go
```diff
@@ -25,41 +25,65 @@ import (
...
 	// AdjudicatorEvents. Those events might be triggered by a Register or
 	// Progress call on the adjudicator from any channel participant.
 	Adjudicator interface {
-		// Register should register the given ledger channel state on-chain.
-		// If the channel has locked funds into sub-channels, the corresponding
-		// signed sub-channel states must be provided.
+		Registerer
+		Withdrawer
+		Progresser
+		EventSubscriber
```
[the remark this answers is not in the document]
manoranjith: `EventSubscriber` removed from `Adjudicator`, as it not used by the client anymore. (It is used only by the `Watcher` and `Watcher` uses the `RegisterSubscriber` method.

watcher/local/watcher_test.go
manoranjith: Moved to `init_test.go`. See PR #213.

watcher/local/watcher.go
```diff
@@ -15,12 +15,92 @@
...
+	if !ok {
+		return nil, nil, errors.New("parent channel not registered with the watcher")
+	}
+	parentCh.subChsAccess.Lock()
+	defer parentCh.subChsAccess.Unlock()
+	return w.startWatching(ctx, parent, signedState)
+}
+
+func (w *Watcher) startWatching(ctx context.Context, parent channel.ID, signedState channel.SignedState) (
+	watcher.StatesPub, watcher.AdjudicatorSub, error) {
```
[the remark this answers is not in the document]
manoranjith: Have updated in PR #213.

watcher/local/watcher.go
manoranjith: It was not need here, this PR is not yet handling the closing of adjudicator subscriptions.

I have added that part in a later commit (concerning the StopWatching functionality. See PR #216.

watcher/local/watcher.go
manoranjith: Just noticed the interface itself need not be public. It is just used to store the pub-sub interfaces in the channel instance (of watcher) to access them while closing.

I have renamed this interface. See PR #216.

watcher/watcher.go
manoranjith: Documenting the method in interfaces was one thing myself and @matthiasgeihs were also discussing  some time ago.

The go's document parser works such that, comment on the interface methods are treated as documentation, it is rather treated as in-line comments for the code.

So, a common practice I noticed in standard library (see [here](https://pkg.go.dev/io#Writer)) was to write documentation on the interfaces rather than methods.

So, we thought we will apply that here.

But, if you see doing otherwise is more beneficial, then we could discuss it and could also change the way we document interfaces.

watcher/local/watcher.go
```diff
@@ -15,12 +15,92 @@
...
+		rs:       cfg.RegisterSubscriber,
+		registry: newRegistry(),
+	}
+	return w, nil
+}
+
+// StartWatchingLedgerChannel starts watching for a ledger channel.
+func (w *Watcher) StartWatchingLedgerChannel(ctx context.Context, signedState channel.SignedState) (
+	watcher.StatesPub, watcher.AdjudicatorSub, error) {
+	return w.startWatching(ctx, channel.Zero, signedState)
```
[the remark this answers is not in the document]
manoranjith: Have updated the implementation to use `nil` instead of `Channel.Zero` See PRs #213 and #214.

watcher/local/registry.go
ggwpez: I will elaborate in #213.

watcher/watcher.go
ggwpez: Hm yes the limitations of channels do not make it easier. Somehow every go framework starts re-inventing subscriptions because there is no idiom for that in go.

watcher/watcher.go
ggwpez: Referenced in #213

watcher/local/watcher.go
ggwpez: moved to #213

watcher/local/watcher.go
ggwpez: moved to #213

watcher/local/watcher.go
ggwpez: Yes i agree, just wanted to let you know that we maybe introduce that function in the future.

watcher/local/registry.go
ggwpez: moved to #213

watcher/local/watcher.go
ggwpez: moved to #213

PR #183, opened 2021-09-09
[pkg/errors] Add Gatherer DoneOrFailed(Ctx)

### LINKED ISSUE #181: pkg/errors.Gatherer: Add `DoneOrFailed()` or similar.
Currently, it's impossible to perform a "Wait for completion of all goroutines or until the first goroutine fails" trivially. We need that behavior in the `erdstall` project. Please either add a function `DoneOrFailed() <-chan error` (or `<-chan struct{}`, I don't really care), or a `WaitDoneOrFailed()` function. While you're at it, it might also be an appropriate time to add context support for the Gatherer, but that's optional.

ggwpez:
Closes #181  
Need this for testing, @RmbRT also seems to appreciate it.

### DIFF
--- a/pkg/errors/gatherer.go
+++ b/pkg/errors/gatherer.go
@@ -20,6 +20,8 @@ import (
 	"sync"
 
 	"github.com/pkg/errors"
+
+	pkgsync "perun.network/go-perun/pkg/sync"
 )
 
 // NewGatherer creates a new error gatherer.
@@ -33,7 +35,7 @@ func NewGatherer() *Gatherer {
 type Gatherer struct {
 	mutex sync.Mutex
 	errs  accumulatedError
-	wg    sync.WaitGroup
+	wg    pkgsync.WaitGroup
 
 	onFails []func()
 	failed  chan struct{} // Closed when an error has occurred.
--- a/pkg/errors/gatherer.go
+++ b/pkg/errors/gatherer.go
@@ -15,6 +15,7 @@
 package errors
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"sync"
@@ -44,6 +45,27 @@ type Gatherer struct {
 // Failed returns a channel that is closed when an error occurs.
 func (g *Gatherer) Failed() <-chan struct{} {
 	return g.failed
+}
+
+// WaitDoneOrFailed returns when either:
+// all routines returned successfully OR
+// any routine returned with an error.
+func (g *Gatherer) WaitDoneOrFailed() {
+	_ = g.WaitDoneOrFailedCtx(context.Background())
+}
+
+// WaitDoneOrFailedCtx returns when either:
+// all routines returned successfully OR
+// any routine returned with an error OR
+// the context was cancelled.
+// The result indicates whether the context was cancelled.
+func (g *Gatherer) WaitDoneOrFailedCtx(ctx context.Context) bool {
+	select {
+	case <-g.failed:
+	case <-g.wg.WaitCh():
+	case <-ctx.Done():
+	}
+	return ctx.Err() == nil
 }
 
 // Add gathers an error. If the supplied error is nil, it is ignored. On the
--- a/pkg/errors/gatherer_test.go
+++ b/pkg/errors/gatherer_test.go
@@ -15,6 +15,7 @@
 package errors_test
 
 import (
+	"context"
 	stderrors "errors"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


pkg/errors/gatherer.go
```diff
@@ -44,6 +47,32 @@ func (g *Gatherer) Failed() <-chan struct{} {
...
+// any routine returned with an error.
+func (g *Gatherer) DoneOrFailed() <-chan struct{} {
+	return g.DoneOrFailedCtx(context.Background())
+}
+
+// DoneOrFailedCtx closes the returned channel when either:
+// all routines returned successfully OR
+// any routine returned with an error OR
+// the context was cancelled.
+func (g *Gatherer) DoneOrFailedCtx(ctx context.Context) <-chan struct{} {
```
RmbRT, line 61: Make it return `bool`, indicating whether it was done or failed within the context deadline. Remove the goroutine. Rename to `WaitDoneOrFailedCtx()`.

pkg/errors/gatherer.go
```diff
@@ -44,6 +47,32 @@ func (g *Gatherer) Failed() <-chan struct{} {
 	return g.failed
 }
 
+// DoneOrFailed closes the returned channel when either:
+// all routines returned successfully OR
+// any routine returned with an error.
+func (g *Gatherer) DoneOrFailed() <-chan struct{} {
```
RmbRT, line 53: Make it return nothing, call the new `WaitDoneOrFailed()` as suggested. Rename to `WaitDoneOrFailed()`.

pkg/errors/gatherer.go
```diff
@@ -44,6 +47,32 @@ func (g *Gatherer) Failed() <-chan struct{} {
...
+		case <-g.failed:
+		case <-g.wg.WaitCh():
+		case <-ctx.Done():
+		}
+		close(ret)
+	}()
+
+	return ret
+}
+
```
RmbRT, line 75: Add `Done() <-chan struct{}` returning the WaitGroup's `WaitCh()`.

pkg/errors/gatherer.go
```diff
@@ -44,6 +47,32 @@ func (g *Gatherer) Failed() <-chan struct{} {
...
+		select {
+		case <-g.failed:
+		case <-g.wg.WaitCh():
+		case <-ctx.Done():
+		}
+		close(ret)
+	}()
+
+	return ret
+}
```
RmbRT, line 74: *(Cannot leave comment on unchanged lines)*
Also add a `WaitCtx(Context) bool` function that calls the waitgroup's `WaitCtx()`.

RmbRT: Might also return `error, bool`, but that's a bit clumsy IMO. Either way is fine with me.

RmbRT (changes requested):

manoranjith:
@ggwpez Where you able to figure out the reason for linter error ?
I am also getting the same error on my PRs as well. But when I try to run the linter locally, I do not get this error.

matthiasgeihs:
> @ggwpez Where you able to figure out the reason for linter error ?
> I am also getting the same error on my PRs as well. But when I try to run the linter locally, I do not get this error.

It looks like the function that is used got deprecated.
https://staticcheck.io/docs/checks#SA1019
Would it be possible to rewrite that code without using the deprecated function?

ggwpez:
> Update: I think we can just modify the panic message:
> `panic(fmt.Sprintf("Expected both interfaces to be zero or both non-zero, got %v, %v", left, right))`

Yes. Its just weird that i cannot locally reproduce the error, just as @manoranjith said.  
I will create a new PR for the fix.

pkg/errors/gatherer.go
ggwpez: done

pkg/errors/gatherer.go
ggwpez: done

pkg/errors/gatherer.go
ggwpez: Would like to not blow up the MR, referenced in https://github.com/perun-network/poly-go/issues/5

pkg/errors/gatherer.go
ggwpez: Would like to not blow up the MR, referenced in https://github.com/perun-network/poly-go/issues/5

RmbRT (approved):

PR #185, opened 2021-09-09
Define smaller interfaces for methods in adjudcator

### LINKED ISSUE #184: Define smaller interfaces for methods in Adjudicator
##### Context
Currently, the adjudicator interface consisted of four methods.

##### Problem
But, only some of the methods were required in any particular place.

#### Possible Solution
We can define an interface for each of the methods and compose the adjudicator interfaces from these smaller interfaces.

manoranjith:
- Previously, the adjudicator interface consisted of four methods.

- But, only some of the methods were required in any particular place.

- Hence, define an interface for each of the methods and compose the
  adjudicator interfaces from these smaller interfaces.

Resolves #184.


channel/adjudicator.go
```diff
@@ -36,30 +36,52 @@ type (
...
+		Registerer
+		Withdrawer
+		Progresser
+		AdjudicatorEventSubscriber
+	}
+
+	// RegistererSubscriber is used the sub-set of Adjudicator interface
+	// required for watching the blockchain for adjudicator events and
+	// registering disputes if required.
+	RegistererSubscriber interface {
```
matthiasgeihs, line 48: Is this used anywhere yet? If not, introduce with the change that will use it?

channel/adjudicator.go
```diff
@@ -36,30 +36,52 @@ type (
 	// AdjudicatorEvents. Those events might be triggered by a Register or
 	// Progress call on the adjudicator from any channel participant.
 	Adjudicator interface {
-		// Register should register the given ledger channel state on-chain.
-		// If the channel has locked funds into sub-channels, the corresponding
-		// signed sub-channel states must be provided.
+		Registerer
+		Withdrawer
+		Progresser
+		AdjudicatorEventSubscriber
```
matthiasgeihs, line 42: Rename to `EventSubscriber` or simply `Subscriber`?

matthiasgeihs (commented):
Some comments on the interface definitions. Let's discuss during our call later.

I have also some comments on the documentation, but we can do that afterwards.

channel/adjudicator.go
manoranjith: Yes, this is used in #182. I think, we `RegistererSubscriber` interface alone to that PR ?

channel/adjudicator.go
manoranjith: Renamed to `EventSubscriber` in this [force push](https://github.com/hyperledger-labs/go-perun/compare/e86f82174c996ed2a88b3d4e9458872444df4563..71f5bc88d52ffd45b33b957d88780009680a6e6c).

channel/adjudicator.go
```diff
@@ -36,30 +36,52 @@ type (
...
-		// Register should register the given ledger channel state on-chain.
-		// If the channel has locked funds into sub-channels, the corresponding
-		// signed sub-channel states must be provided.
+		Registerer
+		Withdrawer
+		Progresser
+		EventSubscriber
+	}
+
+	// RegistererSubscriber is used the sub-set of Adjudicator interface
```
manoranjith, line 45:
```suggestion
	// RegistererSubscriber is a sub-set of Adjudicator interface
```

channel/adjudicator.go
matthiasgeihs: ok

manoranjith:
> Some comments on the interface definitions. Let's discuss during our call later.
> 
> I have also some comments on the documentation, but we can do that afterwards.

@matthiasgeihs As discussed in the call, I have revised the docs comments on the interfaces to idiomatic style (as used in the go standard library for for interfaces like `io.Writer`, `io.WriteSeeker` etc.,)

Also, I have removed the `RegisterSubscriber` method, which can be added as the part of the PR which uses that interface.

matthiasgeihs:
OK, so far this looks fine. However, I would like to wait with approval and merging until we have a complete draft of the implementation of the local watcher. I am not super happy about breaking up the adjudicator interface into many single-method interfaces. It might make sense to only separate those which relevant to the watcher.

PR #187, opened 2021-09-14
Create issue template

matthiasgeihs:
Adding a simple issue template


ggwpez (approved):
I have no experience with these templates, lets see how it works with this first version.

Issue #189, opened 2021-09-15
Channel Backend Test: Generation of inconsistent Params


matthiasgeihs:
### Location

`channel/test/channeltest.go`

### Problem

`func buildModifiedParams` generates a list of modified parameters. However, as their ids are not recomputed, the generated objects have a wrong id.

### Proposal

We create the modified Params such that their id is computed correctly.

PR #192, opened 2021-09-15
190 verify params

### LINKED ISSUE #190: Channel Backend Test: `Verify(addr, params, state, sig)` does not verify params.
### Location

`backend/sim/channel`, `backend/eth/channel`

### Problem

The signature verification method of the channel backend interface expects as input a channel parameters and channel state. However, our two implementations do not verify whether the given parameters are consistent with the given state. In fact, our tests were failing if an implementation is checking and throwing an error in that case. This feels counterintuitive and leads to confusion when implementing backends.

### Proposal

If we provide the params via the interface, we should also verify that their are consistent with the given state. Alternatively, we may consider changing the interface definition.

matthiasgeihs:
Channel backend signature verification: Check Params.

Closes #190 

Needs #189

### DIFF
--- a/channel/test/channeltest.go
+++ b/channel/test/channeltest.go
@@ -96,20 +96,23 @@ func genericVerifyTest(t *testing.T, s *Setup) {
 
 	// Different state and same params
 	ok, err = channel.Verify(addr, s.Params, s.State2, sig)
-	assert.NoError(t, err, "Verify should not return an error")
+	assert.Error(t, err, "Verify should return an error")
 	assert.False(t, ok, "Verify should return false")
 
-	// Different params and same state
-	// -> The backend does not detect this
-
-	// Different params and different state
 	for _, _modParams := range buildModifiedParams(s.Params, s.Params2, s) {
+		// Different params and same state
 		modParams := _modParams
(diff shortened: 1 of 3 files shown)


channel/test/channeltest.go
```diff
@@ -96,20 +96,23 @@ func genericVerifyTest(t *testing.T, s *Setup) {
...
-
-	// Different params and different state
 	for _, _modParams := range buildModifiedParams(s.Params, s.Params2, s) {
+		// Different params and same state
 		modParams := _modParams
+		require.NotEqual(t, *s.Params, modParams)
+		ok, _ = channel.Verify(addr, &modParams, s.State, sig)
+		assert.False(t, ok, "Should not verify with wrong params")
 		for _, _fakeState := range buildModifiedStates(s.State, s.State2, false) {
+			// Different params and different state
```
ggwpez, line 109: I think if you add a require for inequality of the `Params` above, you should also put a  
`require.NotEqual(t, *s.State, fakeState)` here.

ggwpez (commented):
Looks good so far, waiting for the other MR and rebase.

matthiasgeihs:
Closed in favor of #196

PR #194, opened 2021-09-15
193 rename wallettest variable

### LINKED ISSUE #193: wallet/test: AddressBytes != Address.Bytes()
### Location

`wallet/test/wallettest.go`

### Problem

`Setup.AddressBytes` is typically initialized in the corresponding test setups via `Address.Bytes()`, which seems plausible. However, later in the generic wallet tests, it is tested whether `Setup.AddressBytes` can be used with `Address.Decode`, assuming that this always works. In general, this assumption is wrong as there may be good reasons that the result of `Address.Bytes` is not the same as `Address.Encode`, in particular, if we have a variable length encoding.

### Proposal

We rename `Setup.AddressBytes` to `Setup.AddressEncoded`.

matthiasgeihs:
Closes #193 

The variable name was suggesting that it can always be initialized with `Address.Bytes`, but the tests were assuming that the address is decodable, which is not necessarily true for `Address.Bytes`. In particular, note that the length of `Address.Bytes` can be retrieved using the `len` operator, while the length of an encoded address must either be known in advance or contained in the encoding.

### DIFF
--- a/backend/ethereum/wallet/hd/wallet_test.go
+++ b/backend/ethereum/wallet/hd/wallet_test.go
@@ -73,7 +73,7 @@ func TestNewWallet(t *testing.T) {
 func TestSignWithMissingKey(t *testing.T) {
 	setup, accsWallet, _ := newSetup(t, pkgtest.Prng(t))
 
-	missingAddr := common.BytesToAddress(setup.AddressBytes)
+	missingAddr := common.BytesToAddress(setup.AddressEncoded)
 	acc := hd.NewAccountFromEth(accsWallet, accounts.Account{Address: missingAddr})
 	require.NotNil(t, acc)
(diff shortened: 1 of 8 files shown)


ggwpez:
If you want, you can also fix something from our #179 list:  
`wallet.Address needs only to be an Encoder, not Serializer. Decoding is handled by the backend anyway`.  
But i understand if you just want this merged, no need to do it now.  
@matthiasgeihs

matthiasgeihs:
OK, I followed your proposal and replaced `Serializer` with `Encoder`. @ggwpez

ggwpez (approved):
Please rebase @matthiasgeihs, looks good.

matthiasgeihs:
@ggwpez done

ggwpez (approved):

PR #198, opened 2021-09-16
Integrate StatesPub into client

### LINKED ISSUE #197: Integrate statesPub interface into client.
### Location
[client]

### Context

Refactor of watcher for IoT use cases is described in umbrella issue: #172. To integrate the refactored watcher into the client, each channel should publish the off-chain transactions to the watcher via `StatesPub` interface.

```go
type StatesPub interface {
    Publish(channel.Transaction) error
}
```

### Details 

New states are created by the client in two places: when [proposing](https://github.com/hyperledger-labs/go-perun/blob/1fc30ba43eaae379ed5c566bdcf79edafc598b88/client/update.go#L217) and when [accepting](https://github.com/hyperledger-labs/go-perun/blob/1fc30ba43eaae379ed5c566bdcf79edafc598b88/client/update.go#L370). In this two places, client should published the states to the watcher.

However, the `statesPub` instance will be available only after a channel is initialized and the client registers the channel with the watcher. Therefore,
1. Add a field `Publisher` with type `func(channel.Transaction) error` to the `client.Channel` type.
2. Implement a `NoopPublisher` that always returns nil.
3. When initializing the channel, set the `Publisher` to a `NoopPublisher`.
4. Publish the states on the Publisher in these two places: when [proposing](https://github.com/hyperledger-labs/go-perun/blob/1fc30ba43eaae379ed5c566bdcf79edafc598b88/client/update.go#L217) and when [accepting](https://github.com/hyperledger-labs/go-perun/blob/1fc30ba43eaae379ed5c566bdcf79edafc598b88/client/update.go#L370).

In an upcoming pull request, `NoopPublisher` will be replaced by the `Publisher` associated with the watcher when client registers the channel with the watcher.

manoranjith:
Prep step for integrating refactored watcher (for IoT) use cases.

Resolves #197.

Note: Only the last commit is relevant. Other commits are already part of another (PR #182).


matthiasgeihs:
OK, this looks good. Since this PR is very small, you might consider merging it with another PR that also provides a working states pub implementation.

manoranjith:
The implementation of working states pub-sub is included in PR @182. 

I am not sure, if I can extract that into a separate PR. Because, pub-sub implementation and the constructor for the pub-sub are implementation details of the watcher and are not tested separately. Instead, they are tested while testing the local watcher.

Do you think, still I can extract the implementation of states pub-sub (without tests) to a separate PR and merge this PR into that one ? @matthiasgeihs

matthiasgeihs:
@manoranjith I think you mean #182 

What I am missing here is that it is not clear how you would set the `Channel.statesPub` to a working implementation. My guess is that you would do that in `Channel.Watch`? I think it would be helpful if you include that part in this PR as well.

manoranjith:
Yes, I meant `#182`.

Yeah, the second option is possible. In fact, this commit is already part of #205. So, we can close this in  favour of #205.

PR #201, opened 2021-09-16
balance reader interface

### LINKED ISSUE #200: Generic client test setup should not depend on concrete backend type
### Location
client, client/test

### Problem
The struct `RoleSetup` contains property `Backend *ctest.MockBackend`, which is specific to the mock backend setup and therefore cannot be used in other backend tests.

### Proposal
`RoleSetup.Backend` is in fact currently only used for reading account balances. Hence we can create a corresponding interface type `BalanceReader` and use it instead of `*ctest.MockBackend`.

matthiasgeihs:
Closes #200 

Introduces a balance reader interface for the generic client tests.

(Also moves `func makeRoleSetups` to a more suitable location.)

### DIFF
--- a/client/test/role.go
+++ b/client/test/role.go
@@ -29,6 +29,7 @@ import (
 	"perun.network/go-perun/channel/persistence"
 	"perun.network/go-perun/client"
 	"perun.network/go-perun/log"
+	"perun.network/go-perun/wallet"
 	wallettest "perun.network/go-perun/wallet/test"
 	"perun.network/go-perun/wire"
 )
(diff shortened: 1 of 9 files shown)


client/test/role.go
```diff
@@ -54,6 +55,10 @@ type (
 		sync.RWMutex
 	}
 
+	BalanceReader interface {
```
ggwpez, line 58: Its sound pretty self explanatory, but we always comment public stuff in this file.

backend/ethereum/client/init_test.go
```diff
@@ -33,3 +39,20 @@ func init() {
 	// Fix the finality depth for testing.
 	channel.TxFinalityDepth = 3
 }
+
+func makeRoleSetups(s *ethctest.Setup, names [2]string) (setup [2]clienttest.RoleSetup) {
```
ggwpez, line 43: I think `init*.go` files should only contain initialization code.  
This function would better fit into a new package `[backend/eth/client/test] setup.go` since it is generic test code for `[backend/eth/client]`.

ggwpez (approved):

PR #204, opened 2021-09-16
Improve documentation

manoranjith:


pkg/sync/closer.go
```diff
@@ -78,9 +78,9 @@ func (c *Closer) IsClosed() bool {
 	return c.isClosed.IsSet()
 }
 
-// OnClose registers the passed callback to be calle when the Closer is closed.
-// If the Closer is already closed, does nothing. Returns whether the Closer was
-// not yet closed.
+// OnClose registers the passed callback to be called when the Closer is closed.
+// If the Closer is already closed, it does nothing and returns false.
+// If the Closer is not yet closed and call back was registered, returns true.
```
matthiasgeihs, line 83:
```suggestion
// If the Closer is not yet closed, returns true.
```

It always returns true if the Closer is not closed, right?

(Besides that, note that callback should be written as one word.)

pkg/sync/closer.go
```diff
@@ -95,8 +95,9 @@ func (c *Closer) OnClose(handler func()) bool {
 
 // OnCloseAlways registers the passed callback to be called when the Closer is
 // closed.
-// If the Closer is already closed, immediately executes the callback. Returns
-// whether the closer was not yet closed.
+// If the Closer is already closed, it executes the callback immediately and
+// returns false.
+// If the Closer is not yet closed and call back was registered, returns true.
```
matthiasgeihs, line 100: Same comment as above.

pkg/sync/closer.go
manoranjith: Yes, it returns true if the closer is not closed.
Also, changed `call back` -> `callback`.

pkg/sync/closer.go
manoranjith: Updated.

manoranjith:
@matthiasgeihs I have addressed your review comments. 

Force push shows a lot of diffs because, I had rebased this onto latest dev branch.

The diffs concerning this change are shown in the Files Changed section.

pkg/sync/closer.go
matthiasgeihs: OK, so if it always returns true independent of the callback, why do you mention the callback at all? (See my change suggestion above.)

pkg/sync/closer.go
matthiasgeihs: see discussion on line 83 above

matthiasgeihs (commented):
Added comments to the previous discussions. It's just minor, but you extended the comment and I don't get why. I think it would be good if you stated briefly what problem you were having with the previous formulation.

pkg/sync/closer.go
manoranjith: In `OnCloseAlways`, previously it said 
```
If the Closer is already closed, immediately executes the callback
```
It was not clear whether the callback was registered or not in the first read. In the proposed revision, I tried to make it more explicit. But, if you feel, the previous version is already clear, I wouldn't mind dropping this change.

pkg/sync/closer.go
manoranjith: I tried to make it explicit that
1. Callback will be registered only when the channel is not yet closed. It will be ignored otherwise.

matthiasgeihs: OK, but then you should write the comment like that. Currently this is not what the comment says.

PR #205, opened 2021-09-16
Integrate watcher into client

manoranjith:


manoranjith:
Commits older than d9da18b45d8c893369e23a529ac7c3f1310113e1 are already part of another PR, only the commits after that are relevant for this PR.

manoranjith:
Staus: All tests are expected to pass. Race tests fail, to be fixed.

PR #208, opened 2021-09-16
207 fix ensure registered

### LINKED ISSUE #206: Eth channel tests are slow
### Location
backend/eth/channel

### Problem
Since we recently introduced the txFinalityDepth, the eth channel package tests take a relatively long time to complete, approximately 30s on a standard machine.

### Proposal
Decrease the blockInterval to speed up the tests.

### LINKED ISSUE #207: ensureRegistered: ensureRegistered may not detect channel registration
### Location
client/adjudicate.go

### Problem
`func ensureRegistered` currently tries determines whether a channel is registered by checking whether a registered event is seen on the event subscription. However, since our subscription only returns the latest event, it may miss the registered event. In particular, it may see a `ProgressedEvent` or a `ConcludedEvent`, but not the corresponding `RegisteredEvent`, and in such a case will not detect the channel as registered.

(This, by the way, shows a downside of our subscription logic. I think it is worth discussing what are actually the benefits of only returning the latest events. At the moment I rather see disadvantages, namely a more complex implementation and potential problems in usage as described above.)

### Proposal
To ensure that a registered channel is detected even if the `RegisteredEvent` is not the latest event on the subscription, we should additionally check for `ProgressedEvent` and `ConcludedEvent`.

matthiasgeihs:
Closes #207 
Closes #206 

Also fixes another issue where `ensureRegistered` would fail because the channel is registered twice.

### DIFF
--- a/client/adjudicate.go
+++ b/client/adjudicate.go
@@ -417,47 +417,65 @@ func (c *Channel) ensureRegistered(ctx context.Context) error {
 		return nil
 	}
 
-	registeredEvents := make(chan *channel.RegisteredEvent)
+	registered := make(chan error)
+	go func() {
+		registered <- c.awaitRegistered(ctx)
+	}()
 
+	// Register.
+	err := c.registerDispute(ctx)
+	if err != nil {
+		// Only log because channel may already be registered.
+		c.Log().Warnf("registering: %v", err)
+	}
+
+	select {
+	case err = <-registered:
+	case <-ctx.Done():
+		err = ctx.Err()
+	}
+	return err
+}
+
+// awaitRegistered scans for an event indicating that the channel has been
+// registered and waits for the event timeout to elapse.
+func (c *Channel) awaitRegistered(ctx context.Context) error {
 	// Start event subscription.
 	sub, err := c.adjudicator.Subscribe(ctx, c.Params().ID())
 	if err != nil {
 		return errors.WithMessage(err, "subscribing to adjudicator events")
 	}
-	go waitForRegisteredEvent(sub, registeredEvents)
-
-	// Register.
-	err = c.registerDispute(ctx)
-	if err != nil {
-		return errors.WithMessage(err, "registering")
-	}
-
-	select {
-	case e := <-registeredEvents:
-		err = e.Timeout().Wait(ctx)
-		if err != nil {
-			return err
-		}
-	case <-ctx.Done():
-		return ctx.Err()
-	}
-	return nil
-}
-
-// waitForRegisteredEvent waits until a RegisteredEvent has been observed.
-func waitForRegisteredEvent(sub channel.AdjudicatorSubscription, events chan<- *channel.RegisteredEvent) {
 	defer func() {
-		err := sub.Close()
-		if err != nil {
-			log.Warn("Subscription closed with error:", err)
+		if err := sub.Close(); err != nil {
+			c.Log().Warn("Subscription closed with error:", err)
 		}
 	}()
 
-	// Scan for registered event.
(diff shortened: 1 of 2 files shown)


client/adjudicate.go
```diff
@@ -417,23 +417,34 @@ func (c *Channel) ensureRegistered(ctx context.Context) error {
...
-
+	events := make(chan channel.AdjudicatorEvent)
 	// Start event subscription.
 	sub, err := c.adjudicator.Subscribe(ctx, c.Params().ID())
 	if err != nil {
 		return errors.WithMessage(err, "subscribing to adjudicator events")
 	}
-	go waitForRegisteredEvent(sub, registeredEvents)
+	defer func() {
+		if err := sub.Close(); err != nil {
```
ggwpez, line 427: The only error that a `Closer` can return from its `Close` function is an `AlreadyClosedError`.  
Maybe we should either: Rethink the error logic or/and add logging to the `Close` method to not bloat our `defer sub.Close()` calls.

client/adjudicate.go
```diff
@@ -417,23 +417,34 @@ func (c *Channel) ensureRegistered(ctx context.Context) error {
...
+		// Only log because channel may already be registered.
+		c.Log().Warnf("registering: %v", err)
 	}
 
 	select {
-	case e := <-registeredEvents:
+	case e := <-events:
+		// We set phase registered to ensure that we are in the correct phase
+		// even if the channel has been registered by someone else.
+		err = c.setRegisteredRecursive(ctx)
```
ggwpez, line 444: Is this function idempotent? Would it break when there are multiple events received here?

client/adjudicate.go
```diff
@@ -444,20 +455,23 @@ func (c *Channel) ensureRegistered(ctx context.Context) error {
...
-	defer func() {
-		err := sub.Close()
-		if err != nil {
-			log.Warn("Subscription closed with error:", err)
-		}
-	}()
-
-	// Scan for registered event.
+// awaitRegistered scans for an event indicating that the channel has been registered.
+func awaitRegistered(sub channel.AdjudicatorSubscription, events chan<- channel.AdjudicatorEvent) {
```
ggwpez, line 459: You are ignoring any error that `sub` produces, in case it breaks for example.  
`sub.Err()` should be used after `sub.Next()` returned nil.

client/adjudicate.go
matthiasgeihs: I thought the same while looking at my code. This is kind of ugly currently. However, it's a separate issue, hence, let's address it in a separate PR.

client/adjudicate.go
matthiasgeihs: Yes, it's fine to call it twice.

client/adjudicate.go
matthiasgeihs: OK, I refactored the code a bit. Now it also checks `sub.Err` and returns more quickly in the error case.

client/adjudicate.go
ggwpez: referenced in #211

ggwpez (commented):
`ensureRegistered` looks much cleaner now :+1:

ggwpez (dismissed):

ggwpez:
Somehow the [race test](https://github.com/hyperledger-labs/go-perun/pull/208/checks?check_run_id=3634410732#step:5:54) failed in the `setRegisteredRecursive` :thinking:  
I also hope that the shorter timeouts of 10ms work in the CI reliably.  
@matthiasgeihs

ggwpez (approved):

PR #212, opened 2021-09-20
Implement local watcher (1/5): Interface defs and pub-sub implementations

manoranjith:


channel/adjudicator.go
```diff
@@ -25,41 +25,76 @@ import (
...
-		// Register should register the given ledger channel state on-chain.
-		// If the channel has locked funds into sub-channels, the corresponding
-		// signed sub-channel states must be provided.
+		Registerer
+		Withdrawer
+		Progresser
+		EventSubscriber
+	}
+
+	// RegisterSubscriber is the interface that groups Register and Subscribe
```
ggwpez, line 48:
```suggestion
	// RegisterSubscriber is the interface that groups the Register and Subscribe
```

channel/adjudicator.go
```diff
@@ -25,41 +25,76 @@ import (
...
+	// methods.
+	//
+	// These methods are used to watch for adjudicator events on the blockchain
+	// and register disputes, if state in the adjudicator event is not the
```
ggwpez, line 52: Disputes are not registered but states, or just write `and dispute`.

ggwpez: `the state`

watcher/local/adjudicatorpubsub.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Close closes the publisher instance and all the subscriptions associated
+// with it. Any further call to Publish should immediately return error.
+func (a *adjudicatorPubSub) close() error {
+	if a.closed.IsSet() {
+		return errors.New("publisher is closed")
+	}
+	a.err = errors.WithStack(ErrAlreadyClosed)
+	close(a.pipe)
+	a.closed.Set()
```
ggwpez, line 58: Using `IsSet` and `Set` after each other does not prevent a race.  
You can use the `TrySet() bool` function which does both atomically.

watcher/local/adjudicatorpubsub.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+func (a *adjudicatorPubSub) publish(e channel.AdjudicatorEvent) {
+	a.pipe <- e
+}
+
+// Close closes the publisher instance and all the subscriptions associated
+// with it. Any further call to Publish should immediately return error.
+func (a *adjudicatorPubSub) close() error {
```
ggwpez, line 52: If you remove the `error` you can use a `sync.Once`.

watcher/local/adjudicatorpubsub.go
```diff
@@ -0,0 +1,69 @@
...
+	return nil
+}
+
+func (a *adjudicatorPubSub) EventStream() <-chan channel.AdjudicatorEvent {
+	return a.pipe
+}
+
+// Err returns the error after the the subscription is closed.
+func (a *adjudicatorPubSub) Err() error {
+	return a.err
```
ggwpez, line 68: Read is not thread safe.

watcher/local/adjudicatorpubsub.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (a *adjudicatorPubSub) publish(e channel.AdjudicatorEvent) {
+	a.pipe <- e
+}
+
+// Close closes the publisher instance and all the subscriptions associated
+// with it. Any further call to Publish should immediately return error.
+func (a *adjudicatorPubSub) close() error {
+	if a.closed.IsSet() {
+		return errors.New("publisher is closed")
```
ggwpez, line 54: return `a.err` or directly `ErrAlreadyClosed`?

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Close closes the publisher instance and all the subscriptions associated
+// with it. Any further call to Publish should panic.
+func (s *statesPubSub) close() error {
+	if s.closed.IsSet() {
+		return s.err
+	}
+	s.err = errors.WithStack(ErrAlreadyClosed)
+	close(s.pipe)
+	s.closed.Set()
```
ggwpez, line 70: Same comment as above about the `TrySet`.

watcher/watcher.go
```diff
@@ -0,0 +1,75 @@
...
+	// the watcher by calling the StartWatching method.
+	//
+	// After that, it could publish each state on the StatesPub. If any state
+	// is registered or progressed on the blockchain, the corresponding
+	// adjudicator event will be relayed by the watcher on the AdjudicatorSub.
+	// If the registered state is not the latest state (published to the
+	// watcher) then, the watcher will automatically refute by registering the
+	// latest state on the blockchain.
+	//
+	// Once a channel is closed, client can de-register the channel from the
```
ggwpez, line 36:
```suggestion
	// Once a channel is closed, the client can de-register the channel from the
```

watcher/watcher.go
```diff
@@ -0,0 +1,75 @@
...
+	// watcher by calling the StopWatching API. This will close the
```
ggwpez, line 37:
```suggestion
	// watcher by calling the StopWatching function. This will close the
```

watcher/watcher.go
```diff
@@ -0,0 +1,75 @@
...
+	// AdjudicatorSub and StatesSub. Client must not publish states after
```
ggwpez, line 38:
```suggestion
	// AdjudicatorSub and StatesSub. The client must not publish states after
```

watcher/watcher.go
```diff
@@ -0,0 +1,75 @@
...
+	// this.
+	Watcher interface {
+		StartWatchingLedgerChannel(context.Context, channel.SignedState) (StatesPub, AdjudicatorSub, error)
+		StartWatchingSubChannel(context.Context, channel.ID, channel.SignedState) (StatesPub, AdjudicatorSub, error)
```
ggwpez, line 42: The other arguments are pretty self explanatory, but whose channel id is this?  
The root channel or something else?

ggwpez (changes requested):
Looks good overall, just some comments @manoranjith

channel/adjudicator.go
manoranjith: Updated in 8f267f5d

channel/adjudicator.go
manoranjith: Updated in 06a82e12.

watcher/local/adjudicatorpubsub.go
manoranjith: Used `TrySet()`. See 8f267f5d.

watcher/local/adjudicatorpubsub.go
manoranjith: Removed error field in the pub-sub. But, the APIs still return error, attaching the stack at the time of returning error. So, using `atomic.Bool` instead of `sync.Once`.

watcher/local/adjudicatorpubsub.go
manoranjith: Made it thread-safe. See 8f267f5d.

watcher/local/adjudicatorpubsub.go
manoranjith: Updated to directly return `errAlreadyClosed`.

watcher/local/statespubsub.go
manoranjith: Fixed. See 8f267f5d.

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,80 @@
...
+	return nil
+}
+
+func (s *statesPubSub) statesStream() <-chan channel.Transaction {
+	return s.pipe
+}
+
+func (s *statesPubSub) error() error {
+	return s.err
```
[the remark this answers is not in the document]
manoranjith: Fixed. See 8f267f5d.

watcher/watcher.go
manoranjith: Updated. See 22c7d1a9.

watcher/watcher.go
manoranjith: Updated. See 22c7d1a9.

watcher/watcher.go
manoranjith: Updated. See 22c7d1a9.

watcher/watcher.go
manoranjith: It is the parent channel. Yes, it is not self-explanatory, so named that parameter. See 22c7d1a9.

ggwpez (approved):
I forgot to send on commit from yesterday, otherwise its fine for me :+1:

watcher/watcher.go
```diff
@@ -0,0 +1,75 @@
...
+		StopWatching(channel.ID) error
```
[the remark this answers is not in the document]
manoranjith: Yes, I agree.

Though, we might not use the context in the local watcher per-say, this will become for remote watcher implementations.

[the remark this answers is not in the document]
manoranjith: I had added a context.

Since the review is approved, I have squashed all the review commits. See the [diff](https://github.com/hyperledger-labs/go-perun/compare/22c7d1a99fe23c3b7236ae325a6a5e1554e2eb0d..7c6a0fa04c3f6bc6ba571081d7b825f9930849ca) of this force-push.

manoranjith:
Grafted a couple of changes from #213 to this PR as it should have been here.

manoranjith:
@ggwpez Can you re-approve. As the PR approvals are dismissed when we squash and push / rebase and push.

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/pkg/errors"
+
+	"perun.network/go-perun/channel"
+	"perun.network/go-perun/pkg/sync/atomic"
+	"perun.network/go-perun/watcher"
+)
+
+var _ watcher.StatesPub = &statesPubSub{}
+
+// errAlreadyClosed indicates the pub-sub has already been closed.
```
matthiasgeihs, line 29:
```suggestion
// errAlreadyClosed indicates that the pub-sub has already been closed.
```

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// statesSub is used by the watcher to receive transactions.
+	statesSub interface {
+		statesStream() <-chan channel.Transaction
+		error() error
+		close() error
+	}
+)
+
+func newStatesPubSub() *statesPubSub {
```
matthiasgeihs, line 48: Same as above.

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,91 @@
...
+	return &statesPubSub{
+		pipe: make(chan channel.Transaction, statesPubSubBufferSize),
+	}
+}
+
+// Publish publishes the given state to all the subscribers.
```
matthiasgeihs, line 54: You write `all the subscribers`, but I think this is not correct as the implementation doesn't support multiplexing.

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,91 @@
...
+//
+// Each time when a transaction is published, watcher will treat it as the
```
matthiasgeihs, line 56: Which watcher are you referring to?

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,91 @@
...
+// latest transaction without any validation. It is the responsibility of the
+// client to publish transactions in correct order and ensure they are the
```
matthiasgeihs, line 58:
```suggestion
// client to publish transactions in correct order and to ensure that they are
```

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,91 @@
...
+// valid.
+//
+// Context is added as this method implements watcher.StatesPub interface.
+// However, is is not used as there no network communications involved in local
+// watcher.
+//
```
matthiasgeihs, line 64:
```suggestion
```

I think this is self-explanatory. I like that emphasize on code documentation, but sometimes it feels a bit too detailed to me. While documentation is important, it should also be concise. This makes it easier to comprehend and maintain.

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,91 @@
...
+// Returns nil if the state is published. Panics if the subscriptions is
```
matthiasgeihs, line 65:
```suggestion
// Returns nil if the state is published. Panics if the subscription is
```

matthiasgeihs: Can we change the implementation so that it throws an error instead of panicing if the channel is already closed?

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,91 @@
...
+// already closed.
+func (s *statesPubSub) Publish(_ context.Context, tx channel.Transaction) error {
+	s.pipe <- tx
+	return nil
+}
+
+// close closes the publisher instance and all the subscriptions associated
+// with it. Any further call to Publish should panic.
```
matthiasgeihs, line 73: Suggestion: Use sync.Once.Do to close the subscription. The function should not panic. No error return value needed.

Question: Is there a benefit to throwing an error or panicing if you call the function twice?

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	close(s.pipe)
+	return nil
+}
+
+func (s *statesPubSub) statesStream() <-chan channel.Transaction {
+	return s.pipe
+}
+
+func (s *statesPubSub) error() error {
```
matthiasgeihs, line 86: Similar comment as on the other error function. I don't see the point in having it throw an error if the subscription is closed. This is not really an error of the subscription. Since we otherwise don't throw an error, we might just drop the function? If you need a function to know whether the subscription is closed, you could implement a `IsClosed` function.

matthiasgeihs (commented):
A bunch of comments, one major on the interface design, several medium on implementations of close and error functions, some minor on spelling.

channel/adjudicator.go
```diff
@@ -25,41 +25,76 @@ import (
...
+	// a final state can directly be withdrawn after registration.
 	//
 	// Furthermore, an Adjudicator has a method for subscribing to
 	// AdjudicatorEvents. Those events might be triggered by a Register or
 	// Progress call on the adjudicator from any channel participant.
 	Adjudicator interface {
-		// Register should register the given ledger channel state on-chain.
-		// If the channel has locked funds into sub-channels, the corresponding
-		// signed sub-channel states must be provided.
+		Registerer
```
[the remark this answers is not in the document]
manoranjith: I see two aspects to this discussion:

1. Using `Adjudicator` interface in its current form in the `Watcher`.

It is a good practice to require only the smallest possible interface in a function argument. Because, this makes the dependencies of the function explicit.

2. Over time, we might add functionalities like `Progressing` and `Concluding` to the watcher. 

Looking at the [proposal](https://github.com/hyperledger-labs/perun-proposals/blob/main/design/003-IoT-Adoption.md) for adopting perun-framework for IoT use cases, I don't see that watcher will be extended to support these functionalities. On the other hand, this functionalities (Progressing, Concluding) might be implemented by the `Funding` component (shown in the figure 2 in the proposal).

3. Split the methods required only for the watcher.

In that case, the definition of `Adjudicator` interface will be a mix of methods and interfaces. See below:

```
type Adjudicator interface {
    Progress(..)
    Withdraw(...)
    RegisterSubscriber
}

type RegisterSubscriber interface {
    Register(...)
    Subscriber(...{
}
```

This would mean documentation for some methods of the interface would be on the `Adjudicator` type and for the remaining interfaces on the `RegisterSubscriber` type. I prefer to avoid this by defining interfaces for every method and composing the `Adjudicator` from them. But, on this part I wouldn't mind do it as described the above snippet.

watcher/local/adjudicatorpubsub.go
```diff
@@ -0,0 +1,71 @@
...
+
+const adjPubSubBufferSize = 10
+
+type (
+	adjudicatorPubSub struct {
+		closed atomic.Bool
+		pipe   chan channel.AdjudicatorEvent
+	}
+
+	adjudicatorPub interface {
```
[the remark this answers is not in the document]
manoranjith: Yes, I agree that interfaces should be introduced in the PRs using them.

But, in this case, I had created a PR including all the changes for the watcher and then split it up into 5 parts. After splitting it turned out that, earlier PR splits (1/5) has interfaces that will be used in later splits (3/5).

Reorganizing all the commits such that each PR is self contained would require a lot of effort. So, I think if you are okay with it, we ignore the linter warning for unused.

This shouldn't be a problem because, we are merging only into a feature branch, and all the tests, linter checks will pass on the feature branch when all 5 split PRs are merged.

watcher/local/statespubsub.go
```diff
@@ -0,0 +1,91 @@
...
+const statesPubSubBufferSize = 10
+
+type (
+	statesPubSub struct {
+		closed atomic.Bool
+		pipe   chan channel.Transaction
+	}
+
+	// statesSub is used by the watcher to receive transactions.
+	statesSub interface {
```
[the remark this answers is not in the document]
manoranjith: See [this comment](https://github.com/hyperledger-labs/go-perun/pull/212#discussion_r717645347) above.

watcher/local/statespubsub.go
manoranjith: See [this comment](https://github.com/hyperledger-labs/go-perun/pull/212#discussion_r717645347) above.

watcher/local/statespubsub.go
manoranjith: We could do that, but it would come at a cost.

1. We need to use a variable + mutex to co-ordinate the publishing and closing. Because these aspects are done by two different components and should be concurrency-safe.
2. Then, each time we publish, we need to acquire mutex lock, check if the channel is closed and then publish.

I think, it is not worth introducing these complexities in this case, because the close will be called only after client calls stop watching. And it is fair to expect that, the client does not `Publish` after it calls `StopWatching.

watcher/local/statespubsub.go
manoranjith: Yes, the benefit is much simpler implementation of pub-sub. Also, there is no reason for the Publish to be called after channel is closed. See [the above comment.](https://github.com/hyperledger-labs/go-perun/pull/212#discussion_r717660310)

watcher/local/adjudicatorpubsub.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Close closes the publisher instance and all the subscriptions associated
+// with it. Any further call to Publish should immediately return error.
+func (a *adjudicatorPubSub) close() error {
+	if !a.closed.TrySet() {
+		return errors.WithStack(errAlreadyClosed)
+	}
+	close(a.pipe)
+	return nil
+}
```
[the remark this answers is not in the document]
manoranjith: **Short answer:** It possible, I have updated the implementation to use `sync.Once` and updated `close` to return nothing. 

**Explanation:**
As mentioned in the earlier comment with @ggwpez, the reason for using `atomic.Bool` was to return an error when 
1. `close` is called more than once.
2. `Err` method is called after the subscription is closed.

But, thinking through how this error should be handled, I realized
1. Calling `close` multiple times is an acceptable behaviour and the caller only wants to ensure the pub-sub is closed after the call returns. They need not know, if they closed it or it was already closed.

    Hence, there is no need for close to return an error.
2. Calling `Err` after a subscription is closed. In case of local watcher, a subscription is closed only when
    1. Watcher crashes. In that case, client will also have crashed. So error handling is irrelevant.
    2. Client calls `StopWatching` for the channel and the `Watcher` closes the subscription. In this case, we previously returned an `errAlreadyClosed`.
    
    But, I think it is more appropriate to return `nil`, because the subscription was closed on the request of the client and no errors were encountered when closing it. Hence, updated `Err` to always return `nil`.

watcher/local/statespubsub.go
manoranjith: To your suggestion, **short answer** is yes, we can use `sync.Once` and remove error return. I have updated.

**Explanation**:
As, I mentioned in the conversation with @ggwpez earlier, I used `atomic.Bool` to track if a channel is closed and return error in two cases:
1. When `close` is called repeatedly.

    In this case, I realized it is acceptable if close is called multiple times. Only expectation from the caller is that the channel is closed after the `close` call returns. It is relevant if it was closed already or closed as a result of that call.

   Hence, removed error return for close.

2. When `Err` is called after the subscription is closed.

    In this case, the only way a local subscription is closed, is when the client calls `StopWatching` for the channel. After `StopWatching` is called, the `Watcher` will close the subscription. Earlier, we returned a `errAlreadyClosed` error in this case. But, I realized, it is more appropriate to return `nil`, because the subscription was closed only on the request of the client and there was no error encountered while closing it.

    Hence, updated `Err` to always return nil and documented the reason for the same.

watcher/local/statespubsub.go
manoranjith: I agree. I have removed the `error` method. See [the comment](https://github.com/hyperledger-labs/go-perun/pull/212#discussion_r717914622) above for details.

channel/adjudicator.go
```diff
@@ -25,41 +25,76 @@ import (
...
+		EventSubscriber
+	}
+
+	// RegisterSubscriber is the interface that groups the Register and
+	// Subscribe methods.
+	//
+	// These methods are used to watch for adjudicator events on the blockchain
+	// and dispute, if the event is a RegisteredEvent and the state in the
+	// event is not the latest.
+	RegisterSubscriber interface {
```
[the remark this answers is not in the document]
matthiasgeihs: If they use the same naming schema in the standard package, I am fine with that 👍

channel/adjudicator.go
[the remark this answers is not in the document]
matthiasgeihs: OK, I think you have convinced me that we can split the interface up 👍

watcher/local/statespubsub.go
matthiasgeihs: Reading your comment, it seems you would not have implemented my suggestion.

However, reading the code, it looks like you did.

Am I correct? What is the current state?

matthiasgeihs: 👌

matthiasgeihs (commented):
Looks like everything has been addressed. Anything open from your review @ggwpez ?

matthiasgeihs:
@manoranjith Feel free to squash.

PR #222, opened 2021-09-23
Upgrade go to v1.17

### LINKED ISSUE #221: Upgrade go
### Location
go.mod

### Problem
Go 1.17 is available. We may consider upgrading.

### Proposal
Upgrade to go 1.17 in go.mod.

matthiasgeihs:
Closes #221

With go v1.17 the build tag format changes from `+build` to `go:build`.

Also resolves some issues with files using build tags:
- build tag wrap_test was unused and documented as broken
- files race.go and norace.go were using build tags to detect the race flag, but were unused


go.mod
```diff
@@ -12,3 +12,34 @@ require (
 	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 )
+
+require (
```
ggwpez, line 16: Why do we have two `require` blocks now?

matthiasgeihs: This was automatically done by `go mod tidy`. Maybe this is something new in v1.17? You might want to check locally if your `go mod tidy` (v1.17) gives the same results. Note that the second `require` only contains indirect dependencies.

ggwpez: Ok right, i get nearly the same result. Also with two `require` sections…

ggwpez (approved):

PR #223, opened 2021-09-23
Extend generic channel tests

### LINKED ISSUE #219: Extend generic channel tests
### Location
channel/test

### Problem
There do not exist generic tests for the following interfaces: channel.Adjudicator, channel.Funder, channel.AdjudicatorSubscription.
The generic backend test does not cover method `DecodeAsset`.

### Proposal
We add generic tests for channel.Adjudicator, channel.Funder, channel.AdjudicatorSubscription, and channel.Backend.DecodeAsset.

matthiasgeihs:
Closes #219 

Package channel: Adds generic tests for testing funder, adjudicator, and subscription implementation.


backend/ethereum/channel/backend.go
```diff
@@ -263,6 +263,7 @@ func FromEthState(app channel.App, s *adjudicator.ChannelState) channel.State {
 		ID:         s.ChannelID,
 		Version:    s.Version,
 		Allocation: alloc,
+		App:        app,
```
ggwpez, line 266: Created #226

channel/test/adjudicator.go
```diff
@@ -0,0 +1,74 @@
...
+
+	t.Run("progress", func(t *testing.T) {
+		req := a.NewProgressReq(ctx, rng)
+		err := adj.Progress(ctx, *req)
+		require.NoError(t, err, "progressing")
+	})
+
+	t.Run("withdraw", func(t *testing.T) {
+		req, subChannels := a.NewWithdrawReq(ctx, rng)
+		err := adj.Withdraw(ctx, *req, subChannels)
```
ggwpez, line 63: Same comment about having something like a `BalanceReader`.

channel/test/adjudicator.go
```diff
@@ -0,0 +1,74 @@
...
+
+// TestAdjudicator tests the given adjudicator.
+func TestAdjudicator(ctx context.Context, t *testing.T, a Adjudicator) {
+	rng := pkgtest.Prng(t)
+
+	adj := a.Adjudicator()
+
+	t.Run("register and subscribe", func(t *testing.T) {
+		req, subChannels := a.NewRegisterReq(ctx, rng)
+		err := adj.Register(ctx, *req, subChannels)
```
ggwpez, line 48: If the framework had a way to explicitly query the chain with eg. a `DisputeReader`, then it could verify that this worked.

backend/ethereum/channel/adjudicator_test.go
```diff
@@ -33,6 +34,128 @@ import (
...
+func (a *adjudicatorSetup) newAdjudicatorReq(ctx context.Context, rng *rand.Rand, opts ...channeltest.RandomOpt) *channel.AdjudicatorReq {
+	params, state := a.newRandomParamsAndState(rng, opts...)
+
+	// Fund the channel.
+	numParts := len(params.Parts)
+	errs := make(chan error, numParts)
+	for i, funder := range a.setup.Funders {
+		req := channel.NewFundingReq(params, state, channel.Index(i), state.Balances)
+		go func(funder channel.Funder, req channel.FundingReq) {
+			errs <- funder.Fund(ctx, req)
```
ggwpez, line 86: You can use the error gatherer here.

backend/ethereum/channel/adjudicator_test.go
```diff
@@ -33,6 +34,128 @@ import (
...
+
+func (a *adjudicatorSetup) NewRegisterReq(ctx context.Context, rng *rand.Rand) (*channel.AdjudicatorReq, []channel.SignedState) {
+	req := a.newAdjudicatorReq(ctx, rng, channeltest.WithIsFinal(false))
+	return req, nil
+}
+
+func (a *adjudicatorSetup) newAdjudicatorReq(ctx context.Context, rng *rand.Rand, opts ...channeltest.RandomOpt) *channel.AdjudicatorReq {
+	params, state := a.newRandomParamsAndState(rng, opts...)
+
+	// Fund the channel.
```
ggwpez, line 80: I think we could limit the generic test to only make calls on already funded channels.  
This could be done by embedding the `Funder interface` into the `Adjudictor interface` and then let *go-perun* fund the channels before using them.

backend/ethereum/channel/adjudicator_test.go
```diff
@@ -33,6 +34,128 @@ import (
...
+	}
+}
+
+func (a *adjudicatorSetup) newRandomParamsAndState(rng *rand.Rand, opts ...channeltest.RandomOpt) (*channel.Params, *channel.State) {
+	_opts := a.defaultOpts()
+	_opts = append(_opts, opts...)
+	return channeltest.NewRandomParamsAndState(rng, _opts...)
+}
+
+func (a *adjudicatorSetup) defaultOpts() []channeltest.RandomOpt {
```
ggwpez, line 118: Could you not have the `defaultOpts` function as free-standing and pass in the parts and assets?  
Like this you have the same functions one on  the `adjudicatorSetup` and `funderSetup`.

backend/ethereum/channel/adjudicator_test.go
```diff
@@ -33,6 +34,128 @@ import (
...
+}
+
+func (a *adjudicatorSetup) NewWithdrawReq(ctx context.Context, rng *rand.Rand) (*channel.AdjudicatorReq, channel.StateMap) {
+	adj := a.Adjudicator()
+	req := a.newAdjudicatorReq(ctx, rng, channeltest.WithoutApp())
+	subChannels := []channel.SignedState{}
+
+	if !req.Tx.IsFinal {
+		// Register.
+		err := adj.Register(ctx, *req, subChannels)
```
ggwpez, line 140: I think this puts a lot of work on the side of the user of the generic test.  
We normally tried to make it as simple as possible to call these tests.

channel/test/subscription.go
```diff
@@ -0,0 +1,76 @@
...
+	Subscription() channel.AdjudicatorSubscription
+	EmitRegistered(context.Context) (channel.Params, channel.State)
+	EmitProgressed(context.Context) (channel.Params, channel.State)
+	EmitConcluded(context.Context) channel.Params
+}
+
+func TestSubscription(ctx context.Context, t *testing.T, s SubscriptionTestSetup) {
+	sub := s.Subscription()
+
+	{
```
ggwpez, line 41: Will you put `t.Run` calls here or a comment?

channel/test/subscription.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// SubscriptionTestSetup represents a setup for testing a subscription implementation.
+//
+// Subscription should return an instance of the subscription implementation.
+// EmitRegistered should make the subscription emit a registered event.
+// EmitProgressed should make the subscription emit a progressed event.
+// EmitConcluded should make the subscription emit a concluded event.
+type SubscriptionTestSetup interface {
+	Subscription() channel.AdjudicatorSubscription
```
ggwpez, line 32: The instance that is returned here is always the same, or?

channel/test/subscription.go
```diff
@@ -0,0 +1,76 @@
...
+func TestSubscription(ctx context.Context, t *testing.T, s SubscriptionTestSetup) {
+	sub := s.Subscription()
+
+	{
+		params, state := s.EmitRegistered(ctx)
+		e, ok := sub.Next().(*channel.RegisteredEvent)
+		assert.True(t, ok, "registered")
+		assert.True(t, e.ID() == params.ID(), "equal ID")
+		assert.True(t, e.State.Equal(&state) == nil, "equal state")
+		err := e.Timeout().Wait(ctx)
```
ggwpez, line 47: Maybe add a comment to the `EmitRegistered` function that the timeouts should not be picked too large.  
Otherwise if a user will use random timeouts it would block here.

channel/test/subscription.go
```diff
@@ -0,0 +1,76 @@
...
+
+	{
+		params := s.EmitConcluded(ctx)
+		e, ok := sub.Next().(*channel.ConcludedEvent)
+		assert.True(t, ok, "concluded")
+		assert.True(t, e.ID() == params.ID(), "equal ID")
+		err := e.Timeout().Wait(ctx)
+		assert.NoError(t, err, "concluded: waiting")
+	}
+
```
ggwpez, line 69: I think the most important to test would be the event skipping and fast-forwarding of `RegisteredEvent`s.

backend/ethereum/channel/adjudicator_test.go
```diff
@@ -33,6 +34,128 @@ import (
...
+	return &adjudicatorSetup{
+		setup: setup,
+	}
+}
+
+type testAdjudicator struct {
+	*test.SimAdjudicator
+}
+
+// Progress is a no-op because app channels are not supported yet.
```
ggwpez, line 63: It is supported (testable with the `MockApp`), but i think it is ok to do optimistic cases first.

backend/ethereum/channel/subscription_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+type testSubscription struct {
+	channel.AdjudicatorSubscription
+	emitProgressed bool
+	req            *channel.AdjudicatorReq
+}
+
+// Next emulates app channel functionality because app channels are not supported yet.
+func (s *testSubscription) Next() channel.AdjudicatorEvent {
+	if s.emitProgressed {
```
ggwpez, line 80: This assumes that there is only one progression, right?

channel/test/funder.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/stretchr/testify/require"
+	"perun.network/go-perun/channel"
+)
+
+// Funder represents a funder setup used for testing.
+//
+// Funders should return a set of funders used for testing.
+// NewFundingRequests should return a new set of funding requests for the given funders.
+type Funder interface {
```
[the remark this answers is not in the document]
matthiasgeihs: How would you implement the same functionality without using an interface?

channel/test/funder.go
```diff
@@ -0,0 +1,55 @@
...
+	Funders() []channel.Funder
+	NewFundingRequests(context.Context, *testing.T, *rand.Rand) []channel.FundingReq
+}
+
+// TestFunder runs a set of generic funder tests.
+func TestFunder(ctx context.Context, t *testing.T, rng *rand.Rand, f Funder) {
```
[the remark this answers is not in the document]
matthiasgeihs: I think calling the tests generic is redundant because they are located in the generic packages anyways. That is why I decided not to include "generic" in the function name. Also, the naming before "Generic...Test" used before was not idiomatic, because AFAIK usually functions in golang should start with the verb so that they read natural (e.g., "Test...").

channel/test/funder.go
```diff
@@ -0,0 +1,55 @@
...
+	funders := f.Funders()
+	requests := f.NewFundingRequests(ctx, t, rng)
+
+	errs := make(chan error, len(funders))
```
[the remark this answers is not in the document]
matthiasgeihs: To be honest, I find a few things about some of the pkg types confusing. For me they are sometimes not so intuitive to use. Plus, I remember that a design principle of golang is to keep things simple. Hence, I often prefer using standard go functionality, especially when it's rather straightforward.

backend/ethereum/channel/funder_test.go
```diff
@@ -338,6 +338,68 @@ func TestFunder_Fund_multi(t *testing.T) {
...
+func (f *funderSetup) Funders() []channel.Funder {
+	return f.funders
+}
+
+func (f *funderSetup) NewFundingRequests(ctx context.Context, t *testing.T, rng *rand.Rand) []channel.FundingReq {
+	params, state := f.newRandomParamsAndState(rng)
+
+	funders := f.funders
+	numParts := len(funders)
+	requests := make([]channel.FundingReq, numParts)
```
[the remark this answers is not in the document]
matthiasgeihs: Yes.

backend/ethereum/channel/funder_test.go
```diff
@@ -338,6 +338,68 @@ func TestFunder_Fund_multi(t *testing.T) {
...
+}
+
+func (f *funderSetup) newRandomParamsAndState(rng *rand.Rand, opts ...channeltest.RandomOpt) (*channel.Params, *channel.State) {
+	_opts := f.defaultOpts()
+	_opts = append(_opts, opts...)
+	return channeltest.NewRandomParamsAndState(rng, _opts...)
+}
+
+func (f *funderSetup) defaultOpts() []channeltest.RandomOpt {
+	return []channeltest.RandomOpt{
```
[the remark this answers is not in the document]
matthiasgeihs: That's good to know. But does rewriting have benefits in this case?

channel/test/adjudicator.go
```diff
@@ -0,0 +1,74 @@
...
+// NewWithdrawReq should return the inputs to a successful withdraw call.
+type Adjudicator interface {
+	Adjudicator() channel.Adjudicator
+	NewRegisterReq(context.Context, *rand.Rand) (*channel.AdjudicatorReq, []channel.SignedState)
+	NewProgressReq(context.Context, *rand.Rand) *channel.ProgressReq
+	NewWithdrawReq(context.Context, *rand.Rand) (*channel.AdjudicatorReq, channel.StateMap)
+}
+
+// TestAdjudicator tests the given adjudicator.
+func TestAdjudicator(ctx context.Context, t *testing.T, a Adjudicator) {
```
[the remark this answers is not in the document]
matthiasgeihs: Same answer...

channel/test/funder.go
```diff
@@ -0,0 +1,55 @@
...
+	for i := range funders {
+		go func(funder channel.Funder, req channel.FundingReq) {
+			errs <- funder.Fund(ctx, req)
```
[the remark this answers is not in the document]
matthiasgeihs: Definitely, introducing a BalanceReader and checking the Balance after the tests would be good. However, the Eth channel tests don't do that currently and to keep things simple, I kept it likes this at first. Do you prefer to rather do it now?

channel/test/adjudicator.go
```diff
@@ -0,0 +1,74 @@
...
+		req, subChannels := a.NewRegisterReq(ctx, rng)
+		err := adj.Register(ctx, *req, subChannels)
+		require.NoError(t, err, "registering")
+
+		_, err = adj.Subscribe(ctx, req.Params.ID())
+		require.NoError(t, err, "subscribing")
+	})
+
+	t.Run("progress", func(t *testing.T) {
+		req := a.NewProgressReq(ctx, rng)
```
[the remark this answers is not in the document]
matthiasgeihs: That's a good point, and I was considering several options here. One important aspect to consider was that each backend may have its own set of supported features. As you know, some may not support app channels and on-chain progression, for example. Or a backend may only support a certain number of assets of participants in a channel.

The way the tests are written at the moment it is very easy to control these things for the caller of the tests. But if you have some better suggestions, I am happy to discuss in a call.

channel/test/adjudicator.go
matthiasgeihs: Same answer.

channel/test/adjudicator.go
matthiasgeihs: What exactly would you check? How would the implementation look like? I think how disputes are registered is an implementation detail. I can't think of a way currently how to abstract that across all backends. Probably would need a new interface for that, but usually we check the state after registering with a subscription, not via a call to a disputeReader interface.

backend/ethereum/channel/adjudicator_test.go
matthiasgeihs: This comes back to the other discussion. I am aware of the error gatherer, but I find it cleaner to use standard go functionality here.

backend/ethereum/channel/adjudicator_test.go
matthiasgeihs: Yeah, but wouldn't it be somewhat unclean to embed the funder into the adjudicator tests? Maybe for some backends there is another, easier way to set things up. I am not sure, I tried to keep the interfaces and test code simple and minimal at the start, let's discuss in a call.

backend/ethereum/channel/adjudicator_test.go
matthiasgeihs: We may have different default opts for each setup. Also, I like to bind the functions to the calling objects. In this case it's debatable I agree, but I see benefits to both approaches (free standing vs bound).

backend/ethereum/channel/adjudicator_test.go
matthiasgeihs: I agree. However, as pointed out before, this can also have the benefits for the caller to have more control over what features are tested. I am happy if you could make a concrete suggestion on how you would do it.

channel/test/subscription.go
matthiasgeihs: I didn't plan to do so, because the code is dependent on each other. You can't run the later parts without the earlier parts.

channel/test/subscription.go
matthiasgeihs: Yes.

channel/test/subscription.go
matthiasgeihs: But that's obvious? If you pick longer timeouts the tests run longer, and there is usually a test timeout. For the tests to work a developer needs to set things up with suitable parameters.

channel/test/subscription.go
matthiasgeihs: Why?

backend/ethereum/channel/adjudicator_test.go
matthiasgeihs: Good point. Adapted the comment.

backend/ethereum/channel/subscription_test.go
matthiasgeihs: Why do you think it would not work for several progressions? I don't see it.

matthiasgeihs:
After some discussion, marking this as a draft. Planning to resume work on it later.

matthiasgeihs:
Closed until this becomes a higher priority again.

PR #227, opened 2021-09-27
Options to ignore the App + Assets in generic tests.

### LINKED ISSUE #170: Optionalize [channel] GenericBackendTest
**Current situation:** `GenericBackendTest` assumes that the signature of a state depends on all fields.  
The test now fails as we create new backends where the adjudicator does not care about all fields.

**Possible solution:** Add some options like `NoAppData, NoLocked…` and pass them to `GenericBackendTest`.  
The test would then treat those as Don’t-Care.

ggwpez:
Adds `IgnoreApp` and `IgnoreAssets` options to make the `[channel] GenericBackendTest` work with the polkadot backend.

Closes #170


matthiasgeihs (commented):
This looks fine, but I don't know if this is actually needed. Instead of specifying options, you can setup state1 and state2 such that they are equal on the fields that you don't want to test them.

ggwpez:
> This looks fine, but I don't know if this is actually needed. Instead of specifying options, you can setup state1 and state2 such that they are equal on the fields that you don't want to test them.

I will try it without for now and see if it works.

ggwpez:
Confirmed to work with https://github.com/perun-network/perun-polkadot-backend/pull/3, resolving draft status.

matthiasgeihs:
@ggwpez Please see my review above. I am fine with your version, but I have also an alternative suggestion. If you prefer your approach, feel free to merge, but I think you need to rebase first anyways.

ucover:
>Please see my review above. I am fine with your version, but I have also an alternative suggestion. If you prefer your approach, feel free to merge, but I think you need to rebase first anyways.

It would also make it easier to extend, if it turns out that we need to modify it again for a future backend.

Issue #226, opened 2021-09-24
FromEthState unit test


ggwpez:
### Location
`[backend/ethereum/channel] FromEthState`

### Problem
I noticed that the function is not unit-tested after #223 fixed a problem with it it.

### Proposal
Write a test that checks that `FromEthState` works together with `ToEthState` for random states.

PR #228, opened 2021-09-27
backend/eth/channel: Fix state app decoding

matthiasgeihs:
Function `FromEthState` did not set the App correctly on a decoded state.

Relates to #226


ggwpez (approved):
Issue #226 explicitly mentions to add a test for `ToEthState` and `FromEthState`.  
I would then move that to a new issue and merge this as hotfix.

PR #230, opened 2021-09-29
[README] Add read-the-docs badge.

### LINKED ISSUE #229: Ad read-the-docs badge to README
### Location
Top of the `README.md` file.

### Problem
We have the docs linked down below in a section but i think it is better for the visibility when they are on top.  

### Proposal
Add a read-the-docs badge.  
Since we self-host it, it would just be a constant badge image.

ggwpez:
Closes #229


matthiasgeihs (changes requested):
I like the proposal!

However, I find the badge type unsuitable. It is indicating "passing", but it is not a reference to a CI.

Could we use a reference type badge as for example for the go doc reference? That would make it clearer that the badge actually links to the docs themselves, not to some result page for a CI run.

ggwpez:
> I like the proposal!
> 
> However, I find the badge type unsuitable. It is indicating "passing", but it is not a reference to a CI.
> 
> Could we use a reference type badge as for example for the go doc reference? That would make it clearer that the badge actually links to the docs themselves, not to some result page for a CI run.

I also was not really happy with it, but they do not have more here: https://docs.readthedocs.io/en/stable/badges.html  
Searching google does not lead to any other badges. We could also create our own :laughing:   
@matthiasgeihs

matthiasgeihs:
> I also was not really happy with it, but they do not have more here: https://docs.readthedocs.io/en/stable/badges.html Searching google does not lead to any other badges. We could also create our own 😆 @matthiasgeihs

Thanks for the reference. I checked it. It looks like they are integrating the badge differently though: 
https://raw.githubusercontent.com/readthedocs/readthedocs.org/master/README.rst
```
.. |docs| image:: https://readthedocs.org/projects/docs/badge/?version=latest
    :alt: Documentation Status
    :scale: 100%
    :target: https://docs.readthedocs.io/en/latest/?badge=latest
```

ggwpez:
> Thanks for the reference. I checked it. It looks like they are integrating the badge differently though

Yes this is what i meant in the issue with "Since we self-host it, it would just be a constant badge image.".  
We do not use the `readthedocs.com` SAAS to build and host our docs, so AFAIK we cant use their automatic badge system.  
@matthiasgeihs

matthiasgeihs:
OK, I see. In this case we may also create our own badge.

For example:

ggwpez:
> For example: 

Done @matthiasgeihs

Issue #233, opened 2021-10-01
☂️ Create abstractions to make the wire serialization protocol exchangable.


manoranjith:
### Location
[wire], [pkg/io], [wire/net]

### Motivation

Currently, wire messages in go-perun are encoded/decoded using a home-baked. When developing alternate implementations of the perun channel client, in a different programming language, the encoders and decoders have to be re-implemented in that language.

If we are able to exchange the custom format for standard ones like "protocol buffers", the messages can be defined once and the encoders/decoders can be generated for different programming languages. This will eliminate the overhead of implementing and maintaining the encoders/decoders in multiple languages.

However, this it is currently not possible to exchange custom format for a different one, as the necessary abstractions are missing. The scope of this task is to introduce the necessary abstractions.

### Details

Abstractions necessary for making the wire serialization format exchangeable has been discussed and accepted in [proposal#5](https://github.com/hyperledger-labs/perun-proposals/blob/main/design/005-Wire-Encoding-Abstraction.md).

This can be implemented in 6 steps, each involving one PR. This an umbrella issue to track the complete task.

- [x] Step 1
  - [x] Refactor Address,Asset, Data, Action types from (En|De)coder to binary.(Un)marshaler. Discussed in this [comment](https://github.com/hyperledger-labs/go-perun/issues/233#issuecomment-943628627) and tracked in #257.
  - [x] Move out `pkg/io` package to `wire/perun`. #285 
  - [x] Package `pkg/io/test`, which defines generic tests for encoders/decoders is moved to `wire/test` package. #287 
    
- [x] Step 2 #297 

    - [x] Define `EnvelopeSerializer` interface.
    ```
   type EnvelopeSerializer interface {
        Encode(io.Writer, Envelope) error
        Decode(io.Reader) (Envelope, error)
   }
   ```
    - [x] Add a unexported package level variable that holds an instance of `EnvelopeSerializer` and an exported function
    `SetEnvelopeSerializer`, for setting it. The function should panic, if invoked more than once.
    - [x] Implement the `EnvelopeSerializer` for perunio serialization format. The implementation should be placed in new package :`wire/perunio/serializer`. Because, package `wire/perunio` cannot import package `wire`.
    - [x] Add an `init` function in package `wire/perunio/serializer`, which sets the `wire.enveloperSerializer` to perunio envelope serializer. 

[...]

PR #242, opened 2021-10-14
Abstraction of wire serializartion (Draft PR)

manoranjith:
The scope of this PR is to implement changes described in step 1 in #233.

Move the encoding/decoding logic  from package `pkg/io` to `wire/perun and the associated generic test helpers from `pkg/io/test' to `wire/test`.


matthiasgeihs:

PR #249, opened 2021-10-29
✨ Reorg-resistant contract deployment

### LINKED ISSUE #99: Resistant contract deployment
**Location**: `[backend/eth/channel/] deploy.go`

**Problem**: When deploying a contract we currently assume it to be final after one block inclusion (`bind.WaitDeployed`).  
In order to archive reorg resistance, we need to confirm the TX analog to #97.

**Solution**: Create a way to wait for more than one block, in the best case by reusing the code from #97 but possibly with a different finality depth.

matthiasgeihs:
Closes #99

### DIFF
--- a/backend/ethereum/channel/deploy_test.go
+++ b/backend/ethereum/channel/deploy_test.go
@@ -1,0 +1,52 @@
+// Copyright 2021 - See NOTICE file for copyright holders.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package channel_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	ethchannel "perun.network/go-perun/backend/ethereum/channel"
+	"perun.network/go-perun/backend/ethereum/channel/test"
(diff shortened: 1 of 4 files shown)


ggwpez (commented):
Code looks good but where is a test?  
@matthiasgeihs

matthiasgeihs:
Added a test, please have another look @ggwpez

matthiasgeihs:
rebased

backend/ethereum/channel/deploy_test.go
```diff
@@ -0,0 +1,67 @@
...
+				commit <- errors.New("deploy done before commit")
+				return
+			default:
+			}
+			s.SimBackend.Commit()
+		}
+		commit <- nil
+	}()
+
+	select {
```
ggwpez, line 59: This waits for *either* deploy or commit, so it never checks if it actually deployed *and* committed.  
I think the second routine is a bit overkill. You can use this if you want:  
```go
for i := 0; i < int(txFinalityDepth-1); i++ {
	time.Sleep(100 * time.Millisecond) // Give an erroneous deploy function time to finish.
	select {
	case <-deploy:
		t.Error("deployment returned too early")
		t.FailNow()
	default:
	}
	s.SimBackend.Commit()
}
assert.NoError(t, <-deploy)
```
Note: The `-1` should be here as well since `deployContract` already seems to mint one block.  
Putting the sleep into the loop would ensure that the `Commit()` calls have time to be forwarded to the sub before checking for its return.

matthiasgeihs: Thanks for catching this mistake. I refactored several times and apparently didn't do it properly.

I am now following your approach with one exception. Manually reducing the `txFinalityDepth` by 1 to account for the fact that the deploy transaction usually automatically induces one block being mined isn't very clean. Instead, I modified the simBackend to completely allow for manual control. Hence, we can now control that the deployment does not automatically produce a new block and therefore we can use `txFinalityDepth` as we should.

ggwpez: Yes good idea.

ggwpez (approved):

PR #252, opened 2021-11-03
:bug: [client] Start watching for sub-channel

### LINKED ISSUE #251: Start watching does not handle sub-channel correctly
### Location
client/adjudicate.go

### Problem
The watcher interface has a special method for watching a sub-channel, however, this is not used so far by `Channel.Watch`.

### Proposal
Modify `Channel.Watch` to use `StartWatchingSubChannel` for non-ledger channels (i.e., sub-channels and virtual channels).

matthiasgeihs:
Closes #251 

Maybe add tests for watching sub-channels. We only test watching ledger channels.

### DIFF
--- a/client/adjudicate.go
+++ b/client/adjudicate.go
@@ -256,8 +256,8 @@ func (c *Channel) Settle(ctx context.Context, secondary bool) (err error) {
 		return errors.WithMessage(err, "setting phase `Withdrawing` recursive")
 	}
 
-	// Settle.
-	err = c.settle(ctx, secondary)
+	// Withdraw.
+	err = c.withdraw(ctx, secondary)
 	if err != nil {
 		return
 	}
@@ -292,7 +292,7 @@ func (c *Channel) Settle(ctx context.Context, secondary bool) (err error) {
 	return nil
 }
 
-func (c *Channel) settle(ctx context.Context, secondary bool) error {
+func (c *Channel) withdraw(ctx context.Context, secondary bool) error {
 	switch {
 	case c.IsLedgerChannel():
 		subStates, err := c.subChannelStateMap()
--- a/client/adjudicate.go
+++ b/client/adjudicate.go
@@ -71,7 +71,13 @@ func (c *Channel) startWatching() (watcher.StatesPub, watcher.AdjudicatorSub, er
 		State:  currentTx.State.Clone(),
 		Sigs:   currentTx.Sigs,
 	}
-	statesPub, eventsSub, err := c.client.watcher.StartWatchingLedgerChannel(c.Ctx(), signedState)
+
+	statesPub, eventsSub, err := func() (watcher.StatesPub, watcher.AdjudicatorSub, error) {
+		if c.IsLedgerChannel() {
+			return c.client.watcher.StartWatchingLedgerChannel(c.Ctx(), signedState)
+		}
+		return c.client.watcher.StartWatchingSubChannel(c.Ctx(), c.parent.ID(), signedState)
+	}()
 	if err != nil {
 		return nil, nil, errors.WithMessage(err, "registering channel with the watcher")
 	}
--- a/client/init_test.go
+++ b/client/init_test.go
@@ -24,5 +24,5 @@ import (
 // This file initializes the blockchain and logging backend for both, whitebox
 // and blackbox tests (files *_test.go in packages client and client_test).
 func init() {
-	plogrus.Set(logrus.WarnLevel, &logrus.TextFormatter{ForceColors: true})
+	plogrus.Set(logrus.TraceLevel, &logrus.TextFormatter{ForceColors: true})
 }
--- a/client/test/backend.go
+++ b/client/test/backend.go
@@ -96,6 +96,16 @@ func (b *MockBackend) Register(_ context.Context, req channel.AdjudicatorReq, su
 		return nil
 	}
 
+	// Check register requirements.
+	states := make([]*channel.State, 1+len(subChannels))
+	states[0] = req.Tx.State
+	for i, subCh := range subChannels {
+		states[1+i] = subCh.State
+	}
+	if err := b.checkStates(states, checkRegister); err != nil {
+		return err
+	}
+
 	channels := append([]channel.SignedState{
 		{
 			Params: req.Params,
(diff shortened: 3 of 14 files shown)


matthiasgeihs:
Converting to draft, need to check a few things before we can review and merge.

client/init_test.go
```diff
@@ -24,5 +24,5 @@ import (
 // This file initializes the blockchain and logging backend for both, whitebox
 // and blackbox tests (files *_test.go in packages client and client_test).
 func init() {
-	plogrus.Set(logrus.WarnLevel, &logrus.TextFormatter{ForceColors: true})
+	plogrus.Set(logrus.TraceLevel, &logrus.TextFormatter{ForceColors: true})
```
manoranjith, line 27: Did you intended to change the logging level permanently or is it only for testing and you missed to revert ?

client/test/backend.go
```diff
@@ -180,6 +181,10 @@ func (b *MockBackend) Withdraw(_ context.Context, req channel.AdjudicatorReq, su
 	ch := req.Params.ID()
 	if b.isConcluded(ch) {
 		log.Debug("withdraw: already concluded:", ch)
+		e := b.latestEvents[ch]
+		if e.Version() != req.Tx.Version {
+			return fmt.Errorf("Invalid version: expted %v, got %v", e.Version(), req.Tx.Version)
```
manoranjith: Should we change `expted` -> `expected` ?

manoranjith (commented):
@matthiasgeihs  Changes related to fix the bug in Start Watching for sub-channels and the related tests that  were added LGTM. However, I find the naming conventions were different from the one used for the other tests.
Going by the existing convention, I think parts of code for `Tim` and `Susie` could go into  a separate file and `DisputeSusie`, `DisputeTim` would be just `Susie` and `Tim` (just like `Mallory` and `Carol`). What do you think ?

Also, I was not able to clearly follow the other changes (additional checks for withdraw, register). Are they related to this change ? If yes, could you provide some additional context about these changes in the commit message ?

client/adjudicate.go
```diff
@@ -71,7 +71,14 @@ func (c *Channel) startWatching() (watcher.StatesPub, watcher.AdjudicatorSub, er
...
 		Sigs:   currentTx.Sigs,
 	}
-	statesPub, eventsSub, err := c.client.watcher.StartWatchingLedgerChannel(c.Ctx(), signedState)
+
+	statesPub, eventsSub, err := func() (watcher.StatesPub, watcher.AdjudicatorSub, error) {
+		if c.IsLedgerChannel() {
+			return c.client.watcher.StartWatchingLedgerChannel(c.Ctx(), signedState)
+		} else {
+			return c.client.watcher.StartWatchingSubChannel(c.Ctx(), c.parent.ID(), signedState)
+		}
```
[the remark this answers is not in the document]
matthiasgeihs: Using the interface made me think whether it wouldn't be better to just have one interface StartWatching(Context, Channel, Parent)? 🤔 

Anyways, applied your suggestion.

client/init_test.go
matthiasgeihs: Intended, will place it in a separate commit.

client/test/backend.go
matthiasgeihs: Fixed that already. You commented on an outdated commit.

matthiasgeihs:
> @matthiasgeihs Changes related to fix the bug in Start Watching for sub-channels and the related tests that were added LGTM. However, I find the naming conventions were different from the one used for the other tests. Going by the existing convention, I think parts of code for `Tim` and `Susie` could go into a separate file and `DisputeSusie`, `DisputeTim` would be just `Susie` and `Tim` (just like `Mallory` and `Carol`). What do you think ?

We already were using a mixed naming convention. For the role tests, we were using specific first names. The virtual channel test however works differently, for a number of reasons I won't go into here. I found it hard to come up with new names this time, hence the naming I chose. I also think it's more expressive to name the tests by what they do instead of "TestHappySusieTim". We place the code of the two roles in one file together also for the sub-channel happy tests.

> Also, I was not able to clearly follow the other changes (additional checks for withdraw, register). Are they related to this change ? If yes, could you provide some additional context about these changes in the commit message ?

The mock backend has only very limited functionality for testing exactly what we need in the relevant test cases. For making dispute tests work with the mock backend, I had to enhance it's functionality. Before it was mostly used for happy case testing.

matthiasgeihs:
Addressed all comments, still playing around with the timeout in the virtual channel dispute tests. I was hoping that I could decrease it a bit, but doesn't seem to be the case...

matthiasgeihs:
CI stability issues seem actually somewhat unrelated to the timeout in the virtual channel dispute test. At least it is not the virtual channel tests failing anymore.

I also saw that when the ledger channel watcher is shutting down, it sometimes errors with `Error de-registering channel from watcher: cannot de-register when sub-channels are present`.

matthiasgeihs:
Force push:
- Fixed an issue where under certain circumstances an event was written to a closed event channel.
- Fixed an issue where the test expected to receive events for registration of version 0 and version 1, while this may not be true as our event subscription logic may skip outdated events.

matthiasgeihs:
The last commit fixes an issue where a panic in one of the role executions was not propagated and thus resulting in a deadlock because the waitHandler was waiting for the client to be closed.

client/adjudicate.go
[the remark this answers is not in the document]
manoranjith: I thought about it, but then what should be the value of parent when we start watching for ledger channel ?

We may not have avoid using `nil` or equivalent, because, this API could be accessed locally or over remote. In case of accessing over remote, the API call could involve one or two programming languages (client stub & server) and serialization for transporting over network.

client/client_persistence_test.go
```diff
@@ -27,7 +27,7 @@ func TestPersistencePetraRobert(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), twoPartyTestTimeout)
 	defer cancel()
 
-	runTwoPartyTest(ctx, t, func(rng *rand.Rand) (setups []ctest.RoleSetup, roles [2]ctest.Executer) {
+	runAliceBobTest(ctx, t, func(rng *rand.Rand) (setups []ctest.RoleSetup, roles [2]ctest.Executer) {
```
manoranjith, line 30: I think, the new name `runAliceBobTest` is misleading. Because, the tests is actually being run for `Petra` and `Robert` at one place and for `Alice` and `Bob` at another place.

`runTwoPartyTest` or `runTwoPartHappyTest` could be better names for this function.

manoranjith:
@matthiasgeihs LGTM overall. I have two minor suggestions:

1. Naming of `runAliceBobTest`, which I have mentioned above.
2. Reg the commit messages. Could you wrap the body of the commit message at 72 characters ? It is a widely used limit that makes the commit history more readable across different tools.

matthiasgeihs:
Rebase

client/client_persistence_test.go
matthiasgeihs: Reverting this change, because it apparently created other inconsistencies.

manoranjith (approved):
LGTM.

PR #255, opened 2021-11-07
:bug: Prevent confirmNTimes from blocking indefinitely

### LINKED ISSUE #254: 🐛 ContractBackend.confirmNTimes may block indefinitely
### Location
backend/ethereum/channel/contractbackend.go, `func (c *ContractBackend) confirmNTimes` 

### Problem
The function `confirmNTimes` reads the first block header separately in pushes it onto the `heads` channel. However, in some occasions (in particular with very fast block times), the pre-allocated `heads` channel buffer was already filled by the block subscription before the initial head was written. In this case, the code would block indefinitely. See for example [this CI run](https://github.com/hyperledger-labs/go-perun/runs/4130404757?check_suite_focus=true).

### Proposal
Write the initial head before starting the block subscription.

matthiasgeihs:
Closes #254

### DIFF
--- a/backend/ethereum/channel/conclude_test.go
+++ b/backend/ethereum/channel/conclude_test.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"math/rand"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


ggwpez (approved):
Looks good.  
Decreasing the block-time definitely helped to trigger this.

PR #256, opened 2021-11-10
Pkg migration

ggwpez:
- Remove folder `pkg/`
- Replace import paths
- <s>Work-around missing vanity import path, therefore WIP until https://github.com/perun-network/poly-go/issues/1</s>


client/test/doc.go
```diff
@@ -13,4 +13,4 @@
 // limitations under the License.
 
 // Package test contains testing setup types and functions for package client.
-package test // import "perun.network/go-perun/pkg/test"
+package test // import "polycry.pt/poly-go/test"
```
matthiasgeihs, line 16: This looks wrong.

ggwpez: Good catch… seems like the vanity import script is not finding this??

ggwpez: I expanded the vanity script, it now finds it and reports:  
`Package has no or wrong vanity import path: ~/go/src/github.com/perun-network/go-perun/client/test, want: perun.network/go-perun/client/test`.

matthiasgeihs (approved):

PR #266, opened 2021-11-19
Context check

ggwpez:
- Introduces `wire.NewRelayNeverForget()` to remove a context check error
- Fixes one contextcheck error and removes a `nolint:staticcheck`
- Enables contextcheck for non-testing code

The remaining linter warnings in the CI need to be discussed.  
As far as i understand this, it wants us to introduce a context in `Client.Handle` and pass it down to the handlers, aka `handleChannelUpdate`.  
Currently the timeout for the handlers are derived from `client.Ctx` so there could be requests which run until the client is closed.  
We could add some timeouts like `channelUpdateTimeout` akin to `virtualFundingTimeout` and use them in `client.Handle`.  

What do you think?  
@matthiasgeihs

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -28,7 +28,6 @@ linters:
     # These could be enabled in the future:
     - ifshort # we often don't use `if err := …` for readability.
     - tparallel # We don't always use parallel tests.
-    - contextcheck # Wants contexts to be always passed down.
 
     # Deprecated:
     - maligned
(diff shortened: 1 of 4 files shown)


matthiasgeihs:
I don't understand what problem you are addressing and how adding `NewRelayNeverForget` helps with that. Could you elaborate?

ggwpez:
> I don't understand what problem you are addressing and how adding `NewRelayNeverForget` helps with that. Could you elaborate?

The `contextcheck` linter reports 4 warnings less if i introduce the `NewRelayNeverForget` function but i dont get how this should make a difference myself.  
@matthiasgeihs

matthiasgeihs:
OK, just reorganizing the code without deeper reasoning in order to satisfy a linter isn't what we should do.

Either we need to understand why the linter is complaining, and find an appropriate fix. Or if we can't do it (e.g., because it would be a lot of effort) we can suppress the linter warnding.

matthiasgeihs:
Closed in favor of #276

PR #258, opened 2021-11-11
Refactor address and asset types

manoranjith:

PR #271, opened 2021-11-29
:truck: [MIGRATION] Use pkg/io instead of poly-go/io

manoranjith:
- Because, the encoder/decoder implementations are specific to go-perun.

- And the package will be modified in upcoming changes.


manoranjith:
Removed the `tags` `file. 

Regarding placement of the file, I have moved it back into its original location. Moving this file to a more appropriate location, can be done in a separate PR, because we anyways have to re-organize this package and its contents, as a part of  #233.

Also, another open PR (#258) expects this to be in `pkg/io`. If we move this now, then #258 should also be reworked.

matthiasgeihs (approved):

PR #272, opened 2021-11-30
Refactor Data and Action Type

### LINKED ISSUE #257: Refactor encoder/decoder interfaces for abstract types in the core packages of go-perun
### Location
[go-perun] (Most of the packages in go-perun will be impacted).

### Discussed in 
https://github.com/hyperledger-labs/go-perun/issues/233#issuecomment-943628627

### Problem

In go-perun, we implement `perunio.Encoder`/`perunio.Decoder` interfaces for any types that are non-native to go and needs to be sent over the wire. Such non-native types can be grouped into two categories:
1. Struct definitions in core packages. Eg: `channel.Allocation`, `channel.SubAlloc` etc.,
2. Abstract types in core packages. Eg: `wallet.Address`, `channel.App` etc.,

In case of struct definitions, the `Encode`/`Decode` function does only one thing: write/read the message to/from the wire.
But, in case of abstract types, the `Encode`/`Decode` functions do two different things:
1. Convert the blockchain specific concrete data to/from binary format (byte array).
2. Write/Read the data in binary format to/from the wire.

However, these are two different concerns and should be implemented by different functions.

### Proposal

The complete list of affected types are: 
   1. [x] [`wallet.Address`](https://github.com/hyperledger-labs/go-perun/blob/24757c547a921d78f353dd988a8d27827d846ede/wallet/address.go#L29) #258 
   2. [x] [`channel.Asset`](https://github.com/hyperledger-labs/go-perun/blob/24757c547a921d78f353dd988a8d27827d846ede/channel/allocation.go#L97) #258 
   3. [ ] [`channel.App`](https://github.com/hyperledger-labs/go-perun/blob/24757c547a921d78f353dd988a8d27827d846ede/channel/state.go#L54) #272
   4. [ ] [`channel.ActionApp](https://github.com/hyperledger-labs/go-perun/blob/24757c547a921d78f353dd988a8d27827d846ede/channel/app.go#L104) #272

We could apply the changes described below for one type at a time. For time-being the `Bytes` function which does the same thing `encoding.BinaryMarshal`, can be left as-is. In a later issues, we can review its usage and see if it needs to remain or can be removed. 

#### Changes to the abstractions in client, channel, wallet and wire packages.

We could replace the `Encode/Decode` interface on the abstract types with `encoding.BinaryMarshaler`/`encoding.BinaryUnmarshaler` interfaces. These interfaces will then only convert the blockchain specific concrete data to/from binary format (byte array). 

The `Encoding/`Decoding` functions will not be required, because, the `perunio.Encode/Decode` functions are able to work with byte arrays without any additional information.

#### Changes to the wire serialization functions (Encode/Decode) in the pkg/io

To make `io.Encode/Decode`, recognize abstract types and marshal/unmarshal them before/after encoding/decoding,
1. Add a new case in `io.Encode` function: `case encoding.Marshaler` and
2. Add a new case in `io.Decode` function: `case encoding.Unmarshaler`.

#### Changes to the implementations of these abstractions (backend/ethereum, backend/sim)

In each of the backends (sim, ethereum), the implementations of of `Encoder/Decoder` interfaces on the abstract types should be replaced with implementations for `encoding.BinaryMarshaler/encoding.BinaryUnmarshal`.

Cosmwasm backend should also be updated in a separate issue.

manoranjith:
Change Data and Action type from Encoder to binary.(Un)marshaler. 

Relates to #233 (Step 1.2) and closes #257.

### DIFF
--- a/apps/payment/app.go
+++ b/apps/payment/app.go
@@ -16,8 +16,6 @@
 package payment // import "perun.network/go-perun/apps/payment"
 
 import (
-	"io"
-
 	"github.com/pkg/errors"
 
 	"perun.network/go-perun/channel"
@@ -35,9 +33,13 @@ func (a *App) Def() wallet.Address {
 	return a.Addr
 }
 
-// DecodeData does not read anything from the reader and returns new NoData.
-func (a *App) DecodeData(io.Reader) (channel.Data, error) {
-	return Data(), nil
+// NewData returns a new instance of data specific to the payment app,
+// intialized to its zero value.
+//
+// This should be used for unmarshalling the data from its binary
+// representation.
+func (a *App) NewData() channel.Data {
+	return Data()
 }
 
 // ValidTransition checks that money flows only from the actor to the other
--- a/channel/allocation.go
+++ b/channel/allocation.go
@@ -15,7 +15,6 @@
 package channel
 
 import (
-	"bytes"
 	"encoding"
 	"io"
 	"log"
@@ -123,7 +122,7 @@ func NewAllocation(numParts int, assets ...Asset) *Allocation {
 // AssetIndex returns the index of the asset in the allocation.
 func (a *Allocation) AssetIndex(asset Asset) (Index, bool) {
 	for idx, _asset := range a.Assets {
-		if ok, err := equalMarshaling(asset, _asset); ok && err == nil {
+		if ok, err := perunio.EqualBinary(asset, _asset); ok && err == nil {
 			return Index(idx), true
 		}
 	}
@@ -601,7 +600,7 @@ func AssetsAssertEqual(a []Asset, b []Asset) error {
 	}
 
 	for i, asset := range a {
-		if ok, err := equalMarshaling(asset, b[i]); err != nil {
+		if ok, err := perunio.EqualBinary(asset, b[i]); err != nil {
 			return errors.WithMessagef(err, "comparing encoding at index %d", i)
 		} else if !ok {
 			return errors.Errorf("value mismatch at index %d", i)
@@ -609,28 +608,6 @@ func AssetsAssertEqual(a []Asset, b []Asset) error {
 	}
 
 	return nil
-}
-
-func equalMarshaling(a, b encoding.BinaryMarshaler) (bool, error) {
-	// golang does not have a XOR
-	if (a == nil) != (b == nil) {
-		return false, errors.New("only one argument was nil")
-	}
-	// just using a == b would be too easy here since go panics
-	if (a == nil) && (b == nil) {
-		return true, nil
-	}
-
-	encodedA, err := a.MarshalBinary()
-	if err != nil {
-		return false, errors.Wrap(err, "EqualEncoding encode error")
-	}
-	encodedB, err := b.MarshalBinary()
-	if err != nil {
-		return false, errors.Wrap(err, "EqualEncoding encode error")
-	}
-
-	return bytes.Equal(encodedA, encodedB), nil
 }
 
 var _ perunio.Serializer = new(SubAlloc)
--- a/channel/app.go
+++ b/channel/app.go
@@ -163,6 +163,33 @@ func (d OptAppDec) Decode(r io.Reader) (err error) {
 	return errors.WithMessage(err, "resolve app")
 }
 
+// OptAppAndDataEnc makes an optional pair of App definition and Data encodable.
+type OptAppAndDataEnc struct {
+	App  App
+	Data Data
+}
+
+// Encode encodes an optional pair of App definition and Data.
+func (o OptAppAndDataEnc) Encode(w io.Writer) error {
+	return perunio.Encode(w, OptAppEnc{App: o.App}, o.Data)
+}
+
+// OptAppAndDataDec makes an optional pair of App definition and Data decodable.
+type OptAppAndDataDec struct {
+	App  *App
+	Data *Data
+}
+
+// Decode decodes an optional pair of App definition and Data.
+func (o OptAppAndDataDec) Decode(r io.Reader) (err error) {
+	if err = perunio.Decode(r, OptAppDec{App: o.App}); err != nil {
+		return err
+	}
+
+	*o.Data = (*o.App).NewData()
+	return perunio.Decode(r, *o.Data)
+}
+
 // AppShouldEqual compares two Apps for equality.
 func AppShouldEqual(expected, actual App) error {
 	if IsNoApp(expected) && IsNoApp(actual) {
--- a/channel/app.go
+++ b/channel/app.go
@@ -15,6 +15,7 @@
 package channel
 
 import (
+	"encoding"
 	"io"
 
 	"github.com/pkg/errors"
@@ -34,11 +35,15 @@ type (
 		// returns false.
 		Def() wallet.Address
 
-		// DecodeData decodes data specific to this application. This has to be
-		// defined on an application-level because every app can have completely
-		// different data; during decoding the application needs to be known to
-		// know how to decode the data.
-		DecodeData(io.Reader) (Data, error)
+		// NewData returns a new instance of data specific to NoApp, intialized
+		// to its zero value.
+		//
+		// This should be used for unmarshalling the data from its binary
+		// representation.
+		//
+		// This has to be defined on an application-level because every app can
+		// have completely different data.
+		NewData() Data
 	}
 
 	// A StateApp is advanced by full state updates. The validity of state
@@ -90,18 +95,24 @@ type (
 		// parameters and version must be 0.
 		InitState(*Params, []Action) (Allocation, Data, error)
 
-		// DecodeAction decodes actions specific to this application. This has to be
-		// defined on an application-level because every app can have completely
-		// different actions; during decoding the application needs to be known to
-		// know how to decode an action.
-		DecodeAction(io.Reader) (Action, error)
+		// NewAction returns an instance of action specific to this
+		// application. This has to be defined on an application-level because
+		// every app can have completely different action. This instance should
+		// be used for unmarshalling the binary representation of the action.
+		//
+		// It is zero-value is safe to use.
+		NewAction() Action
 	}
 
 	// An Action is applied to a channel state to result in new state.
-	// Actions need to be Encoders so they can be sent over the wire.
-	// Decoding happens with ActionApp.DecodeAction() since the app context needs
-	// to be known at the point of decoding.
-	Action = perunio.Encoder
+	//
+	// It is sent as binary data over the wire. For unmarshalling an action from
+	// its binary representation, use ActionApp.NewAction() to create an Action
+	// instance specific to the application and then unmarshal using it.
+	Action interface {
+		encoding.BinaryMarshaler
+		encoding.BinaryUnmarshaler
+	}
 
 	// AppResolver provides functionality to create an App from an Address.
 	// The AppResolver needs to be implemented for every state channel application.
--- a/channel/mock_app.go
+++ b/channel/mock_app.go
@@ -15,11 +15,11 @@
 package channel
 
 import (
-	"io"
+	"encoding/binary"
+	"fmt"
 
 	"github.com/pkg/errors"
 
-	perunio "perun.network/go-perun/pkg/io"
 	"perun.network/go-perun/wallet"
 )
 
@@ -37,9 +37,16 @@ var (
 // MockOp serves as Action and State.Data for MockApp.
 type MockOp uint64
 
+// mockOpLen is the length of binary representation of mockOp, in bytes.
+const mockOpLen = 8
+
 var (
 	_ Action = (*MockOp)(nil)
 	_ Data   = (*MockOp)(nil)
+
+	// byteOrder used for marshalling/unmarshalling mockOp to/from its binary
+	// representation.
+	byteOrder = binary.BigEndian
 )
 
 const (
@@ -61,14 +68,21 @@ func NewMockOp(op MockOp) *MockOp {
 	return &op
 }
 
-// Encode encodes a MockOp into an io.Writer.
-func (o MockOp) Encode(w io.Writer) error {
-	return perunio.Encode(w, uint64(o))
+// MarshalBinary marshals MockOp to its binary representation.
+func (o MockOp) MarshalBinary() ([]byte, error) {
+	data := make([]byte, mockOpLen)
+	byteOrder.PutUint64(data, uint64(o))
+	return data, nil
 }
 
-// Decode decodes a MockOp from an io.Reader.
-func (o *MockOp) Decode(r io.Reader) error {
-	return perunio.Decode(r, (*uint64)(o))
+// UnmarshalBinary unmarshals MockOp from its binary representation.
+func (o *MockOp) UnmarshalBinary(data []byte) error {
+	if len(data) != mockOpLen {
+		//nolint: goerr113  // We do not want to define this as constant error.
+		return fmt.Errorf("unexpected length %d, want %d", len(data), mockOpLen)
+	}
+	*o = MockOp(byteOrder.Uint64(data))
+	return nil
 }
 
 // Clone returns a deep copy of a MockOp.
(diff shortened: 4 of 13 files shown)


channel/allocation.go
```diff
@@ -123,7 +122,7 @@ func NewAllocation(numParts int, assets ...Asset) *Allocation {
 // AssetIndex returns the index of the asset in the allocation.
 func (a *Allocation) AssetIndex(asset Asset) (Index, bool) {
 	for idx, _asset := range a.Assets {
-		if ok, err := equalMarshaling(asset, _asset); ok && err == nil {
+		if ok, err := perunio.EqualBinary(asset, _asset); ok && err == nil {
```
matthiasgeihs, line 125: We are currently working on a #279  that introduces the Asset type with an Equal method. This should be then used here, depending on which we merge first.

channel/allocation.go
```diff
@@ -601,7 +600,7 @@ func AssetsAssertEqual(a []Asset, b []Asset) error {
 	}
 
 	for i, asset := range a {
-		if ok, err := equalMarshaling(asset, b[i]); err != nil {
+		if ok, err := perunio.EqualBinary(asset, b[i]); err != nil {
```
matthiasgeihs, line 603: We are currently working on a #279  that introduces the Asset type with an Equal method. This should be then used here, depending on which we merge first.

channel/app.go
```diff
@@ -34,11 +35,13 @@ type (
...
-		// defined on an application-level because every app can have completely
-		// different data; during decoding the application needs to be known to
-		// know how to decode the data.
-		DecodeData(io.Reader) (Data, error)
+		// NewData returns an instance of data specific to this application.
+		// This has to be defined on an application-level because every app can
+		// have completely different data. This instance should be used for
+		// unmarshalling the binary representation of the data.
+		//
+		// It is zero-value is safe to use.
```
matthiasgeihs, line 43:
```suggestion
		// It is zero-value and is safe to use.
```
Is that what you wanted to say?

channel/app.go
```diff
@@ -90,18 +93,24 @@ type (
...
-		// defined on an application-level because every app can have completely
-		// different actions; during decoding the application needs to be known to
-		// know how to decode an action.
-		DecodeAction(io.Reader) (Action, error)
+		// NewAction returns an instance of action specific to this
+		// application. This has to be defined on an application-level because
+		// every app can have completely different action. This instance should
+		// be used for unmarshalling the binary representation of the action.
+		//
+		// It is zero-value is safe to use.
```
matthiasgeihs, line 101:
```suggestion
		// It is zero-value and safe to use.
```
Like that?

channel/app.go
```diff
@@ -163,6 +172,33 @@ func (d OptAppDec) Decode(r io.Reader) (err error) {
 	return errors.WithMessage(err, "resolve app")
 }
 
+// OptAppAndDataEnc makes an optional pair of App definition and Data encodable.
+type OptAppAndDataEnc struct {
```
matthiasgeihs, line 176: Why differentiate between `OptAppAndDataEnc` and `OptAppAndDataDec`? Which aspect of `OptAppAndDataEnc` is optional?

channel/app.go
```diff
@@ -163,6 +172,33 @@ func (d OptAppDec) Decode(r io.Reader) (err error) {
...
+
+// OptAppAndDataDec makes an optional pair of App definition and Data decodable.
+type OptAppAndDataDec struct {
+	App  *App
+	Data *Data
+}
+
+// Decode decodes an optional pair of App definition and Data.
+func (o OptAppAndDataDec) Decode(r io.Reader) (err error) {
+	if err = perunio.Decode(r, OptAppDec{App: o.App}); err != nil {
```
matthiasgeihs, line 194: Where is the optional character of this decoding method?

channel/app.go
```diff
@@ -163,6 +172,33 @@ func (d OptAppDec) Decode(r io.Reader) (err error) {
...
+		return err
+	}
+
+	*o.Data = (*o.App).NewData()
+	return perunio.Decode(r, *o.Data)
```
matthiasgeihs, line 199: Why do you pass by value?

channel/mock_app.go
```diff
@@ -61,14 +62,21 @@ func NewMockOp(op MockOp) *MockOp {
 	return &op
 }
 
-// Encode encodes a MockOp into an io.Writer.
-func (o MockOp) Encode(w io.Writer) error {
-	return perunio.Encode(w, uint64(o))
+// MarshalBinary marshals MockOp to its binary representation.
+func (o MockOp) MarshalBinary() ([]byte, error) {
+	data := make([]byte, mockOpBinaryLength)
+	binary.PutUvarint(data, uint64(o))
```
matthiasgeihs, line 68: Use `binary.BigEndian.PutUint64` instead?

channel/mock_app.go
```diff
@@ -61,14 +62,21 @@ func NewMockOp(op MockOp) *MockOp {
...
+	return data, nil
 }
 
-// Decode decodes a MockOp from an io.Reader.
-func (o *MockOp) Decode(r io.Reader) error {
-	return perunio.Decode(r, (*uint64)(o))
+// UnmarshalBinary unmarshals MockOp from its binary representation.
+func (o *MockOp) UnmarshalBinary(data []byte) error {
+	mockOp, cntBytesRead := binary.Uvarint(data)
```
matthiasgeihs, line 74: Use `binary.BigEndian.Uint64` instead?

matthiasgeihs: Then potentially adapt below.

channel/noapp.go
```diff
@@ -64,16 +62,19 @@ func (noApp) ValidInit(_ *Params, s *State) error {
...
 // IsNoData returns whether an app data is NoData.
 func IsNoData(d Data) bool {
-	_, ok := d.(noData)
+	_, ok := d.(*noData)
 	return ok
 }
 
-// Encode does nothing.
-func (noData) Encode(io.Writer) error { return nil }
+// MarshalBinary does nothing and always returns and empty byte array and nil.
```
matthiasgeihs, line 73:
```suggestion
// MarshalBinary does nothing and always returns an empty byte array and nil.
```

channel/state.go
```diff
@@ -50,9 +51,14 @@ type (
...
 	// Data is the data of the application running in this app channel.
-	// Decoding happens with App.DecodeData.
+	//
+	// It is sent as binary data over the wire. For unmarshaling a data from
+	// binary representation, use App.NewData() to create an instance of Data
+	// specific to this application and then unmarshal using it.
 	Data interface {
-		perunio.Encoder
+		encoding.BinaryMarshaler
+		encoding.BinaryUnmarshaler
```
matthiasgeihs, line 60: Why need `Unmarshaler` on interface when we didn't need `Decoder` before?

channel/state.go
```diff
@@ -104,20 +110,16 @@ func (s *State) Clone() *State {
 
 // Encode encodes a state into an `io.Writer` or returns an `error`.
 func (s State) Encode(w io.Writer) error {
-	err := perunio.Encode(w, s.ID, s.Version, s.Allocation, s.IsFinal, OptAppEnc{s.App}, s.Data)
-	return errors.WithMessage(err, "state encode")
+	return errors.WithMessage(
+		perunio.Encode(w, s.ID, s.Version, s.Allocation, s.IsFinal, OptAppAndDataEnc{s.App, s.Data}),
```
matthiasgeihs, line 114: Why not use same struct for encoding and decoding?

client/proposalmsgs.go
```diff
@@ -187,9 +187,9 @@ func (p BaseChannelProposal) NumPeers() int {
 
 // Encode encodes the BaseChannelProposal into an io.Writer.
 func (p BaseChannelProposal) Encode(w io.Writer) error {
-	return perunio.Encode(w,
-		p.ChallengeDuration, p.NonceShare,
-		OptAppAndDataEnc{p.App, p.InitData}, p.InitBals, p.FundingAgreement)
+	optAppAndDataEnc := channel.OptAppAndDataEnc{App: p.App, Data: p.InitData}
```
matthiasgeihs, line 190: Why not use same struct for encoding and decoding?

pkg/io/equal_binary.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"bytes"
+	"encoding"
+
+	"github.com/pkg/errors"
+)
+
+// EqualBinary returns whether the binary representation of the two values `a` and `b` are equal.
+// It returns an error when marshalling fails.
+func EqualBinary(a, b encoding.BinaryMarshaler) (bool, error) {
```
matthiasgeihs, line 26: Still needed when we have `Asset.Equal`?

pkg/io/serialize.go
```diff
@@ -31,7 +31,7 @@ var byteOrder = binary.LittleEndian
 
 // Encode encodes multiple primitive values into a writer.
 // All passed values must be copies, not references.
-func Encode(writer io.Writer, values ...interface{}) (err error) {
+func Encode(writer io.Writer, values ...interface{}) (err error) { //nolint: cyclop
```
matthiasgeihs, line 34: Usually try to avoid `//nolint`. If you need it, add a comment why. Why you need it here?

pkg/io/serialize.go
```diff
@@ -62,6 +62,10 @@ func Encode(writer io.Writer, values ...interface{}) (err error) {
 				return errors.WithMessage(err, "writing length of marshalled data")
 			}
 
+			// Nothing to be encoded when length is zero.
+			if length == 0 {
+				break
+			}
 			err = ByteSlice(data).Encode(writer)
```
matthiasgeihs, line 69: This is a rare case where we still have an `Encode` method. Would it make sense to replace that with a new method name?

pkg/io/serialize.go
```diff
@@ -62,6 +62,10 @@ func Encode(writer io.Writer, values ...interface{}) (err error) {
 				return errors.WithMessage(err, "writing length of marshalled data")
 			}
 
+			// Nothing to be encoded when length is zero.
+			if length == 0 {
```
matthiasgeihs, line 66: Couldn't it be that the serializer needs to write something to ensure that an object has been "registered"?

pkg/io/serialize.go
```diff
@@ -108,6 +112,10 @@ func Decode(reader io.Reader, values ...interface{}) (err error) {
 				return errors.WithMessage(err, "reading length of binary data")
 			}
 
+			// Nothing to be decoded when length is zero.
+			if length == 0 {
```
matthiasgeihs, line 116: Maybe sometimes we want to unmarshal the empty string?

channel/app.go
manoranjith: Because Data itself is an interface and interfaces internally store reference to the actual data.

Effectively, we are passing the reference to the underlying data as value.

channel/state.go
manoranjith: Because, earlier it was decoded by calling `App.DecodeData`, which instantiated an instance of Data and then decoded using it.

Now, `App` has method `NewData`, which just instantiates an instance of Data, which implements `encoding.BinaryUnmarshaler`. This Data can then be directly decoded by passing it to the `pkg.Decode` function.

Essentially, this change moves all notions of decoding to `pkg/io` package.

channel/state.go
manoranjith: It can be done. But, I have not changed it as the scope of this PR is only to convert Data, Action from Encoders to binary Data type. 

Since, this PR in itself is a bit large, my suggestion would be to do that change in another PR. What do you think ?

client/proposalmsgs.go
manoranjith: See https://github.com/hyperledger-labs/go-perun/pull/272#discussion_r763695117

pkg/io/equal_binary.go
manoranjith: Yes, because when validating a state we use it to compare `Data`.

However, in a separate PR, we could implement `Data.Equal` and then remove this function.

pkg/io/serialize.go
manoranjith: We can still recognize that its an empty data, because we encode/decode the length of data.

manoranjith: But, I do not understand your question. What is the case, where we unmarshal the empty string ?

pkg/io/serialize.go
manoranjith: Yes, we do write the length of data (which is zero). We just don't write the data itself. Because, there is nothing to write.

pkg/io/serialize.go
manoranjith: Are you referring to the `Encode` method on the `ByteSlice` type ?

channel/allocation.go
manoranjith: Yes, we could replace this instance with `Asset.Equal` method.

However, regarding which PR to merge first, I would suggest we do this one before #279. Because, though we don't require `equalMarshalling` function for comparing two assets, we still need it for comparing two data. See https://github.com/hyperledger-labs/go-perun/pull/272#discussion_r763697134

channel/allocation.go
manoranjith: https://github.com/hyperledger-labs/go-perun/pull/272#discussion_r763705784

channel/app.go
manoranjith: The pair of "data" and "app" together are optional.

This was an existing construct, which I only moved it from one package to another. If we need to modify this, I think it is beyond the scope of this PR and we should do in a separate one.

channel/app.go
manoranjith: See https://github.com/hyperledger-labs/go-perun/pull/272#discussion_r763713098

pkg/io/serialize.go
manoranjith: It is needed here because, the Encode function by design has a long switch case with one case for each switch statement, which cause the cyclomatic complexity to exceed the limits in our linter.

Added a comment in 464caf77.

channel/noapp.go
manoranjith: Fixed in 11a98044

channel/mock_app.go
manoranjith: Updated in cba3884e.

channel/mock_app.go
manoranjith: Yes, I think constant length encoding is better, as it is simpler to replicate across implementations. Updated in 0dc3cca7.

channel/app.go
manoranjith: I have rephrased it in 27f6382f . Does it look fine now ?

channel/app.go
manoranjith: I have rephrased it in 27f6382f . Does it look fine now ?

apps/payment/app.go
```diff
@@ -35,9 +33,9 @@ func (a *App) Def() wallet.Address {
 	return a.Addr
 }
 
-// DecodeData does not read anything from the reader and returns new NoData.
-func (a *App) DecodeData(io.Reader) (channel.Data, error) {
-	return Data(), nil
+// NewData returns an instance of noData.
```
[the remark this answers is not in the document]
manoranjith: Updated in 17b5b4d9.

pkg/io/serialize.go
matthiasgeihs: I thought. Maybe the serializer sometimes want to write something on the stream to signal that there was an empty string / byte array written. For example, just write uint16(0) as the length without any data following afterwards. Not sure if that is relevant here though.

pkg/io/serialize.go
matthiasgeihs: I am referring to `ByteSlice.Encode`. Is that terminology still suitable and consistent with the changes that we did?

pkg/io/serialize.go
manoranjith: Okay, I get it now. In this particular serializer implementation, it is not relevant, because it does not write anything.

manoranjith: @matthiasgeihs

pkg/io/serialize.go
manoranjith: Yes, it is still consistent. Because, in case of binary types, we first Encode/Decode their length and then Encode/Decode them as ByteSlices.

We do so because, in case of binary types, the core abstraction which calls Encode/Decode does not know the length of data, as it could be different for different backends.

While, in case of ByteSlices, if the core abstraction is Encoding/Decoding them, it does know the length and hence the length need not be Encoded.

pkg/io/serialize.go
matthiasgeihs: OK, I see, and a different serializer implementation could handle it differently.

channel/app.go
matthiasgeihs: Created an issue #281

pkg/io/equal_binary.go
matthiasgeihs: Created #280

matthiasgeihs (dismissed):

matthiasgeihs (approved):

PR #276, opened 2021-12-03
Enable context check linter

### LINKED ISSUE #275: Enable context check linter
### Problem
We enabled a bunch of linter #260 , but didn't enable the context check linter because of some more intricate usages of contexts in go-perun.

### Proposal
Enable the context linter in a separate PR and resolve the previously encountered issues.

matthiasgeihs:
Closes #275

### DIFF
--- a/client/virtual_channel.go
+++ b/client/virtual_channel.go
@@ -90,7 +90,7 @@ func (c *Client) handleVirtualChannelFundingProposal(
 ) {
 	err := c.validateVirtualChannelFundingProposal(ch, prop)
 	if err != nil {
-		c.rejectProposal(responder, err.Error())
+		c.rejectProposal(responder, err.Error()) //nolint:contextcheck
 	}
 
 	ctx, cancel := context.WithTimeout(c.Ctx(), virtualFundingTimeout)
@@ -98,10 +98,10 @@ func (c *Client) handleVirtualChannelFundingProposal(
 
 	err = c.fundingWatcher.Await(ctx, prop)
 	if err != nil {
-		c.rejectProposal(responder, err.Error())
+		c.rejectProposal(responder, err.Error()) //nolint:contextcheck
 	}
 
-	c.acceptProposal(responder)
+	c.acceptProposal(responder) //nolint:contextcheck
 }
 
 func (c *Channel) watchVirtual() error {
@@ -348,7 +348,7 @@ func (c *Client) matchFundingProposal(ctx context.Context, a, b interface{}) boo
 	}
 
 	go func() {
-		err := virtual.watchVirtual()
+		err := virtual.watchVirtual() //nolint:contextcheck // The context will be derived from the channel context.
 		c.log.Debugf("channel %v: watcher stopped: %v", virtual.ID(), err)
 	}()
 	return true
--- a/watcher/local/watcher.go
+++ b/watcher/local/watcher.go
@@ -195,7 +195,7 @@ func (w *Watcher) startWatching(
 		Sigs:  signedState.Sigs,
 	}
 	ch.Go(func() { ch.handleStatesFromClient(initialTx) })
-	ch.Go(func() { ch.handleEventsFromChain(w.rs, w.registry) })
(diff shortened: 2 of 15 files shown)


wire/relay.go
```diff
@@ -64,16 +66,24 @@ func (p *Relay) Close() error {
...
 	if p.IsClosed() {
 		return
 	}
 
-	p.cache.Cache(ctx, predicate)
+	p.cache.Cache(predicate)
+}
+
+// ReleaseCache disable caching for the given predicate.
+func (p *Relay) ReleaseCache(predicate *Predicate) {
```
manoranjith, line 82: Should we add the release method to the `wire.Cache` interface ? Because, I think the behavior of the Cache interface has changed.

Earlier, the caching of version zero updates was skipped when the context passed to `Accept` has been cancelled.
Now, contexts are not used for disabling cache. Instead, an explicit `Release` call is used.

It looks like a better approach to me. Also, could you document this change in the commit message ?

manoranjith (changes requested):

client/virtual_channel.go
```diff
@@ -348,7 +348,7 @@ func (c *Client) matchFundingProposal(ctx context.Context, a, b interface{}) boo
 	}
 
 	go func() {
-		err := virtual.watchVirtual()
+		err := virtual.watchVirtual() //nolint:contextcheck
```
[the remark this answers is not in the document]
matthiasgeihs: Created #282

watcher/local/watcher.go
```diff
@@ -355,15 +355,13 @@ func (ch *ch) handleEventsFromChain(registerer channel.Registerer, chRegistry *r
...
+// each of its children. It then registers a dispute for the channel tree.
 //
 // This function assumes the callers has locked the parent channel.
 func registerDispute(r *registry, registerer channel.Registerer, parentCh *ch) error {
 	parentTx, subStates := retreiveLatestSubStates(r, parentCh)
 
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-	err := registerer.Register(ctx, makeAdjudicatorReq(parentCh.params, parentTx), subStates)
+	err := registerer.Register(context.TODO(), makeAdjudicatorReq(parentCh.params, parentTx), subStates)
```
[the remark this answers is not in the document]
matthiasgeihs: Created #282

wire/relay.go
matthiasgeihs: `Cache` is not an interface. You mean `Cacher`? However, `Cacher` by its name is intended to only deal with caching, not releasing. I think the terminology might be mixed up now. It might make sense to refactor this a bit. I am not even sure if there is any use in having the interface, as we only use one implementation and this part is also not exposed to the outside as far as I can see.

If you have a concrete suggestion on what to change (maybe remove the interface?) feel free to create an issue for that. But I think we should rather do it in a separate PR, if you do such a change at all.

manoranjith: Yes I meant the `Cacher` interface. I thought we should include `Release` method in the `Cacher` interface because it complements the `Cache` method, by disabling caching of a particular item. Looking further, I noticed
1. `Cacher` interface is referenced in the client package. Here, we use only the `Cache` method and not the `Release` method. 
2. `Cache` is the only implementation of `Cacher` interface. The `Release` method on it is an implementation detail, that is referenced only inside the wire package.

 
So, my suggestion would be that, in a separate PR:
1. We can make `Cache`, its methods and the `MakeCache` function unexported.
2. We can retain the `Cacher` interface, because it allows us to freely refactor the `Cache` implementation (like removing use of context and adding Release method, which you did in this PR).

manoranjith:
LGTM. Can you rebase onto the tip of base branch and squash the review commits ?

wire/relay.go
matthiasgeihs: Created #286

matthiasgeihs:
Rebased and squashed.

manoranjith (approved):

matthiasgeihs:
rebased

manoranjith (approved):

PR #277, opened 2021-12-03
Lint comments on exported objects (functions, types, ...)

### LINKED ISSUE #267: Linter does not report undocumented publics
### Location
`.golangci.yml`

### Problem
The [default suppression](https://golangci-lint.run/usage/configuration/) filter of golangci-lint removes the "should have comment or be unexported" warnings.  
Therefore public types/functions without comment currently pass the CI unnoticed.  

### Proposal
This can be disabled with `exclude-use-default: false` but would then require us to copy the remaining defaults that we still want to use.  
Just copying some of the default values does not seem to work so well either, since some are deprecated…

matthiasgeihs:
Closes #267

### DIFF
--- a/backend/sim/wallet/address.go
+++ b/backend/sim/wallet/address.go
@@ -60,7 +60,7 @@ func (a *Address) Bytes() []byte {
 	return data[:]
 }
 
-// ByteArray converts an address into a 64-byte array. The returned array
+// byteArray converts an address into a 64-byte array. The returned array
 // consists of two 32-byte chunks representing the public key's X and Y values.
 func (a *Address) byteArray() (data [addrLen]byte) {
 	xb := a.X.Bytes()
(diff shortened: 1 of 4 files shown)


matthiasgeihs:
Close and reopen to trigger DCO.

backend/sim/wallet/address.go
```diff
@@ -63,13 +62,13 @@ func (a *Address) Bytes() []byte {
 
 // ByteArray converts an address into a 64-byte array. The returned array
```
manoranjith, line 63: byteArray....

matthiasgeihs: Not directly related to my changes, but shows that the linter only cares about exported functions...

matthiasgeihs:
Addressed your comment.

manoranjith (approved):
LGTM.

PR #279, opened 2021-12-03
Introduce and use Asset.Equal

### LINKED ISSUE #247: Make Asset comparable
I think we should extend the `Asset` interface (not in this MR).  
`EqualEncoding` was a work-around from the beginning.

_Originally posted by @ggwpez in https://github.com/hyperledger-labs/go-perun/pull/240#discussion_r737525058_

matthiasgeihs:
Closes #247

### DIFF
--- a/backend/ethereum/channel/asset.go
+++ b/backend/ethereum/channel/asset.go
@@ -32,7 +32,28 @@ import (
 )
 
 // Asset is an Ethereum asset.
-type Asset = wallet.Address
+type Asset struct {
+	wallet.Address
+}
+
+// NewAssetFromAddress creates a new asset from an Ethereum address.
+func NewAssetFromAddress(a common.Address) *Asset {
+	return &Asset{*wallet.AsWalletAddr(a)}
+}
+
+// EthAddress returns the Ethereum address representation of the asset.
+func (a Asset) EthAddress() common.Address {
+	return common.Address(a.Address)
+}
+
+// Equal returns true iff the asset equals the given asset.
+func (a Asset) Equal(b channel.Asset) bool {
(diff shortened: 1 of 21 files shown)


matthiasgeihs:
rebase

backend/ethereum/channel/asset.go
```diff
@@ -32,7 +32,28 @@ import (
 )
 
 // Asset is an Ethereum asset.
-type Asset = wallet.Address
```
manoranjith, line 35: How about defining it as `type Asset walllet.Address` ?

The advantage in this case would be that, we wouldn't need `NewAssetFromAddress` and `EthAddress` methods.
The previously used ways to covert to and from wallet.Address or common.Address will continue to work.

manoranjith (changes requested):
Have one suggestion regarding definition of the `Asset` type.

backend/ethereum/channel/asset.go
matthiasgeihs: It's possible, but in this case methods like `MarshalBinary` wouldn't be inherited and therefore we would have to reimplement them. I don't see why `NewAssetFromAddress` and `EthAddress` then. How would you create a new Asset with your approach? How would you get the underlying `common.Address`. The options I can see appear quite lengthy.

manoranjith: On how to get the EthAddress, the approach would be what it was before the change `ethwallet.Address(Asset)`.  But, as you have pointed out, it is a bit verbose to use.

However, since the `MarshalBinary` function wouldn't be inherited, the current approach seems better.

manoranjith (approved):
LGTM

PR #284, opened 2021-12-08
Use fixed length format for (un)marshalling asset

### LINKED ISSUE #283: Use `binary.BigEndian.PutUint64` for sim.Asset
### Location

[backend/sim/asset]

### Problem

Currently, we use `perunio.Encode` method to encode the asset. This couples the `sim.Asset` to a specific serialization implementation.

### Proposal

As suggested by @matthiasgeihs  in the [review of 272](https://github.com/hyperledger-labs/go-perun/pull/272#discussion_r763154010_), use `binary.BigEndian.PutUint64` directly for marshalling the `sim.Asset` which is of type int64.

manoranjith:
- Previously, pkgio.(Encode|Decode) methods were used for
  marshallign/unmarshalling the Asset. This coupled the asset to a
  specific serialization implementation (pkg/io).

- Instead, use binary.BigEndian.Putuint64 directly.

Closes #283

### DIFF
--- a/backend/sim/channel/asset.go
+++ b/backend/sim/channel/asset.go
@@ -15,16 +15,19 @@
 package channel
 
 import (
-	"bytes"
+	"encoding/binary"
 	"fmt"
 	"math/rand"
 
 	"perun.network/go-perun/channel"
-	"perun.network/go-perun/wire/perunio"
 )
 
-// AssetBinaryLen is the length of binary representation of asset, in bytes.
-const AssetBinaryLen = 8
+// assetLen is the length of binary representation of asset, in bytes.
+const assetLen = 8
+
+// byteOrder used for marshalling/unmarshalling asset to/from its binary
+// representation.
+var byteOrder = binary.BigEndian
 
 // Asset simulates a `channel.Asset` by only containing an `ID`.
 type Asset struct {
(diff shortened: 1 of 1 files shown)


backend/sim/channel/asset.go
```diff
@@ -40,17 +43,17 @@ func NewRandomAsset(rng *rand.Rand) *Asset {
...
+	return data, nil
 }
 
 // UnmarshalBinary unmarshals the asset from its binary representation.
 func (a *Asset) UnmarshalBinary(data []byte) error {
-	if len(data) != AssetBinaryLen {
-		return fmt.Errorf("unexpected address length %d, want %d", len(data), AssetBinaryLen) //nolint: goerr113
+	if len(data) != assetLen {
+		//nolint: goerr113  // We do not want to define this as constant error.
+		return fmt.Errorf("unexpected length %d, want %d", len(data), assetLen)
```
matthiasgeihs, line 55: `//nolint` should be used on the line where it is relevant. (Also typically no space between '//nolint:' and the specified linter name, see [here](https://golangci-lint.run/usage/false-positives/).)
```suggestion
		return fmt.Errorf("unexpected length %d, want %d", len(data), assetLen) //nolint:goerr113  // We do not want to define this as constant error.
```

manoranjith: I had put it above the line, because it was making the line very long. But, I am okay with either of the approaches.

matthiasgeihs (approved):

PR #287, opened 2021-12-09
Merge perunio/test into wire/test

manoranjith:
Completes step 1.3 described in #233.

This is done in two steps:

1. In the first step, the test helper function `TestMsg` which runs a set of generic tests on Msg is moved to perunio/test, where the other generic tests are located.

2. In the second step, the package "perunio/test" which contains test helpers for running various kinds of generic tests is merged into wire/test.

Details of the individual changes have been documented in the commit message.


wire/perunio/test/msgtest.go
```diff
@@ -12,29 +12,30 @@
...
+	msg.Msg, err = wire.Decode(reader)
 	return err
 }
 
-// TestMsg performs generic tests on a wire.Msg object.
-func TestMsg(t *testing.T, msg Msg) {
-	test.GenericSerializerTest(t, &serializerMsg{msg})
+// MsgSerializerTest performs generic serializer tests on a wire.Msg object.
+func MsgSerializerTest(t *testing.T, msg wire.Msg) {
+	t.Helper()
```
matthiasgeihs: This is not a helper.

wire/perunio/test/msgtest.go
```diff
@@ -12,29 +12,30 @@
...
-	msg.Msg, err = Decode(reader)
+	msg.Msg, err = wire.Decode(reader)
 	return err
 }
 
-// TestMsg performs generic tests on a wire.Msg object.
-func TestMsg(t *testing.T, msg Msg) {
-	test.GenericSerializerTest(t, &serializerMsg{msg})
+// MsgSerializerTest performs generic serializer tests on a wire.Msg object.
+func MsgSerializerTest(t *testing.T, msg wire.Msg) {
```
matthiasgeihs: Should be called `TestMsgSerialization`.

client/proposalmsgs_test.go
```diff
@@ -70,7 +71,7 @@ func TestChannelProposalReqSerialization(t *testing.T) {
 			m, err = clienttest.NewRandomVirtualChannelProposal(rng, client.WithNonceFrom(rng), app)
 			require.NoError(t, err)
 		}
-		wire.TestMsg(t, m)
+		iotest.MsgSerializerTest(t, m)
```
matthiasgeihs: It's difficult to review if you change the same line several times within a single PR.

wire/test/msgtest.go
```diff
@@ -12,29 +12,30 @@
...
-	msg.Msg, err = Decode(reader)
+	msg.Msg, err = wire.Decode(reader)
 	return err
 }
 
-// TestMsg performs generic tests on a wire.Msg object.
-func TestMsg(t *testing.T, msg Msg) {
-	test.GenericSerializerTest(t, &serializerMsg{msg})
+// MsgSerializerTest performs generic serializer tests on a wire.Msg object.
+func MsgSerializerTest(t *testing.T, msg wire.Msg) {
```
matthiasgeihs, line 38:
```suggestion
func TestMsgSerialization(t *testing.T, msg wire.Msg) {
```

wire/test/msgtest.go
```diff
@@ -12,29 +12,30 @@
...
+	t.Helper()
```
matthiasgeihs, line 39:
```suggestion
```

wire/cache_internal_test.go
```diff
@@ -15,14 +15,25 @@
...
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	wtest "perun.network/go-perun/wallet/test"
 	"polycry.pt/poly-go/test"
 )
 
+// NewRandomEnvelope - copy from wire/test for internal tests.
```
matthiasgeihs, line 28:
```suggestion
// NewRandomEnvelope - copy from wire/test for internal tests.
```

This comment looks like a leftover. Either just explain what the function does or remove.

wire/cache_internal_test.go
```diff
@@ -15,14 +15,25 @@
...
+func NewRandomEnvelope(rng *rand.Rand, m Msg) *Envelope {
```
matthiasgeihs, line 29: If that function is only used for internal testing, it should be unexported.
```suggestion
func newRandomEnvelope(rng *rand.Rand, m Msg) *Envelope {
```

wire/test/msgtest.go
```diff
@@ -12,29 +12,30 @@
 // See the License for the specific language governing permissions and
```
matthiasgeihs, line 12: Since "test" is now in the pathname, you can rename the file to just "msg.go"?

wire/test/msgtest.go
manoranjith: My understanding is that, this is a helper function that runs serialization tests for any implementation of the interface `Msg`. In the same way as `GenericSerializerTest` or `GenericMarshalerTest`. Hence, I followed the same conventions for those two functions and

1. Named it this way.
2. Put it in a file name appended with test.
3. Marked it as helper.

Is this not the case ?

wire/perunio/test/msgtest.go
manoranjith: See https://github.com/hyperledger-labs/go-perun/pull/287#discussion_r765649069

wire/perunio/test/msgtest.go
manoranjith: See https://github.com/hyperledger-labs/go-perun/pull/287#discussion_r765649069

wire/test/msgtest.go
manoranjith: See https://github.com/hyperledger-labs/go-perun/pull/287#discussion_r765649069

wire/test/msgtest.go
manoranjith: See https://github.com/hyperledger-labs/go-perun/pull/287#discussion_r765649069

wire/cache_internal_test.go
manoranjith: The comment isn't a left over. The function is originally defined in `wire/test`. Because, `wire` cannot import `wire/test`, it is duplicated here.

wire/cache_internal_test.go
manoranjith: Yes, it could be unexported.

manoranjith: https://github.com/hyperledger-labs/go-perun/pull/287/commits/b3e6fbc57aa4b13221fcbdb5758e9c61291211ff

wire/test/msgtest.go
matthiasgeihs: **Regarding Helper**
[Helper doc](https://pkg.go.dev/testing#T.Helper): Helper marks the calling function as a test helper function. When printing file and line information, that function will be skipped. Helper may be called simultaneously from multiple goroutines.

I think this is not what we want here.

**Regarding the function name**
Before it was called `TestMsg`. The point is our naming scheme is not consistent anyway. And according to go guidelines function names should naturally read well as if you would read a sentence, which `MsgSerializerTest` does not, but `TestMsgSerialization ` does.

**Regarding the file name**
You didn't give a reason specifically why you appended "test" to the file name, so I can only restate that I think it's not needed as the package is already called "test".

manoranjith: On marking this as helper, we do not want the test runner to report this line, rather the actual line where this is being called.
For instance, see the below example, where I intentionally failed the tests:

1. When we mark it as a helper function, the runner reports failure in the actual tests (proposalmsgs_test.go:170).

```
❯ got -run TestChannelProposalAccSerialization/ledger_channel
pkg/test: using rootSeed 1639049305959782908
--- FAIL: TestChannelProposalAccSerialization (0.00s)
    --- FAIL: TestChannelProposalAccSerialization/ledger_channel (0.00s)
        proposalmsgs_test.go:170:
                Error Trace:    serializertest.go:31
                                                        msgtest.go:40
                                                        proposalmsgs_test.go:170
                Error:          Not equal:
                                .........
                Test:           TestChannelProposalAccSerialization/ledger_channel
FAIL
exit status 1
FAIL    perun.network/go-perun/client   0.010s                                            /0.4s
```

2. When we don't mark it as helper, it reports the test is in `msgtest.go:40`, which is false, because the actual test case is elsewhere.
```
❯ got -run TestChannelProposalAccSerialization/ledger_channel
pkg/test: using rootSeed 1639049311486872753
--- FAIL: TestChannelProposalAccSerialization (0.00s)
    --- FAIL: TestChannelProposalAccSerialization/ledger_channel (0.00s)
        msgtest.go:40:
                Error Trace:    serializertest.go:31
                                                        msgtest.go:40
                                                        proposalmsgs_test.go:170
                Error:          Not equal:
                                .......
                Test:           TestChannelProposalAccSerialization/ledger_channel
FAIL
exit status 1
FAIL    perun.network/go-perun/client   0.010s                                            /0.4s
```

However, in both cases, the complete path will be shown in the Error Trace field.

manoranjith: Regarding the name, if we being a function with `Test...` it means its a test function. But, in our case it is only a helper to run a set of generic tests. Also, the function would be read as `test.TestMsgSerialization`, which I think would cause a stutter.

manoranjith: Regarding the filename, the package contains two kinds of things:
1. Things that can be used in tests. These are placed in  `randomizer.go` and `serializinglocalbus.go`.
2. Helper functions that run generic tests. These are placed in `bustest.go`, `marshalertest.go` and `serializertest.go`.

Hence, I think we should name it `msgtest.go`, as it falls in the second category.

matthiasgeihs: **Regarding Helper**
OK, your example convinced me. I thought it was then not printing the line where the error occurred. 👍 

**Regarding the function name**
OK.

**Regarding the file name**
OK.

wire/cache_internal_test.go
matthiasgeihs: The point is that you are making a reference to an external package. I don't think it's very useful. You could try to place that function somewhere where both locations can use it. Anyways, up to you.

matthiasgeihs (commented):
LGTM, some open comments but up to you.

:refactor: is not a gitmoji. You mean `:recycle:`?

Want to squash before merge?

client/proposalmsgs_test.go
manoranjith: I understand, I will try to minimize this next time.

wire/cache_internal_test.go
manoranjith: The only place to place it where both packages could use would be `wire` package. But, that will pollute public API of `wire` package. So, I think this is an optimal middle ground.

manoranjith:
Yeah, it was supposed to be `:recycle:` Yes, I will also squash the review commits.

manoranjith:
Squashed, rebased, all tests are passing and ready to be merged. @matthiasgeihs

matthiasgeihs (approved):

PR #289, opened 2021-12-09
💥 Rename `Channel.Update`

### LINKED ISSUE #288: :boom: Client: Rename Channel.UpdateBy -> Channel.Update
### Location

client/update.go

### Problem

At some point we introduced `Channel.UpdateBy` as a safe method for updating a channel. The previous method `Channel.Update` exists but is rarely used. Having the two methods may cause confusion.

### Proposal

Remove `Channel.Update` in favor of `Channel.UpdateBy` and rename `Channel.UpdateBy` to `Channel.Update`.

matthiasgeihs:
Closes #288

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ func main() {
 	}()
 
 	// send a channel update request to the other channel peer(s)
-	err = ch.UpdateBy(ctx, func(s *channel.State) error {
+	err = ch.Update(ctx, func(s *channel.State) error {
 		// update state s, e.g., moving funds or changing app data
 	})
 	if err != nil { /* handle error */ }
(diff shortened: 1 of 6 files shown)


manoranjith (commented):
LGTM.

One minor suggestion would be that, in the body of the first commit (db300fb190a2b4e303bd049565adc2fbb2e10cb5), two points have been written as a single para. 
```
Channel.UpdateBy should be used instead. Also remove
`TestChannel_Update_NilArgs` as it was not working as intended and does
not seem very useful overall.
```
Alternatively, it could be written as bullet points:
```
- Channel.UpdateBy should be used instead.
- Also remove `TestChannel_Update_NilArgs` as it was not working as intended
  and does not seem very useful overall.
```
```

manoranjith (approved):

PR #291, opened 2021-12-09
Client: Adapt ForceUpdate to Update

### LINKED ISSUE #290: Client: Make Channel.Update and ForceUpdate have the same function signature 💥
### Location

client/update.go, client/adjudicate.go

### Problem

`Channel.Update` and `Channel.ForceUpdate` are functionally very similar but their function signature is slightly different. While `Channel.Update` accepts a state modification function that returns an error, the state modification received by `Channel.ForceUpdate` does not have this feature. It's not clear why we have that difference.

### Proposal

We adapt the function signature of `Channel.ForceUpdate` to the one of `Channel.Update`.

matthiasgeihs:
Resolves #290

### DIFF
--- a/client/adjudicate.go
+++ b/client/adjudicate.go
@@ -178,13 +178,16 @@ func (c *Channel) registerDispute(ctx context.Context) error {
 	return nil
 }
 
-// ForceUpdate enforces a channel update on the adjudicator.
+// ForceUpdate enforces a channel update on the adjudicator. `update` gets as
+// input a copy of the current state and is expected to modify the state as
+// intended. It should not to increment the version number as this is
+// automatically.
 //
 // Returns TxTimedoutError when the program times out waiting for a transaction
-// to be mined.
-// Returns ChainNotReachableError if the connection to the blockchain network
-// fails when sending a transaction to / reading from the blockchain.
-func (c *Channel) ForceUpdate(ctx context.Context, update func(*channel.State)) error {
+// to be mined. Returns ChainNotReachableError if the connection to the
+// blockchain network fails when sending a transaction to / reading from the
+// blockchain.
+func (c *Channel) ForceUpdate(ctx context.Context, update func(*channel.State) error) error {
 	err := c.ensureRegistered(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


client/adjudicate.go
```diff
@@ -200,9 +203,12 @@ func (c *Channel) ForceUpdate(ctx context.Context, update func(*channel.State))
 	ar := c.machine.AdjudicatorReq()
 
 	// Update state
-	state := c.machine.State().Clone()
+	state := ar.Tx.State.Clone()
```
manoranjith, line 206: It was a bit hard for me to understand why `ar.Tx.State.Clone()` is used over `c.machine.State().Clone()`. Only after looking into `c.machine,AdjudicatorReq()`, I understood they were the same.

Because, we are locking the machine mutex and the `currentTx` cannot change,
I prefer we use the previous version ``c.machine.State().Clone()`. and move the definition of `ar`, just before the line where it is required.
Then the function would read as
1...... 
2. Fetch the latest state and update it.
3. .... 
4. Generate adjudicator request and make on-chain call.

manoranjith (commented):
Have one minor suggestion.

Also, I think you should rebase onto `main` after #289 is merged ? @matthiasgeihs

client/adjudicate.go
matthiasgeihs: That doesn't work because the adjudicator request must contain the previous state.

matthiasgeihs:
> Have one minor suggestion.
> 
> Also, I think you should rebase onto `main` after #289 is merged ? @matthiasgeihs

Rebased. I think you suggestion didn't apply as intended, see https://github.com/hyperledger-labs/go-perun/pull/291#discussion_r766531048.

client/adjudicate.go
manoranjith: Okay. Now, I understand why this change makes sense. And I read it as
1. Generate adjudicator request from the current state.
2. Update the state in the adjudicator request.
3. Send both the old and the new states to the contract.

manoranjith (approved):
LGTM.

PR #293, opened 2021-12-14
Define an abstract type for wallet.Sig

### LINKED ISSUE #292: Add an abstract type for wallet.Sig
### Location
[wallet] 

### Problem
Currently `wallet.Sig` is defined as a variable size byte array in the abstract `wallet` package. However, the length of the signature depends upon the signature scheme that we use, which in-turn depends upon the wallet backend we use. Hence, in order to `Encode/Decode` a `wallet.Sig`, we directly use the `wallet.Sig.Encode` / wallet.Backend.DecodeSig` functions.

The problem is that, these `Encode/DecodeSig` functions do two different things:

    1. Convert the blockchain specific concrete data to/from binary format (byte array).
    2. Write/Read the data in binary format to/from the wire.

### Proposal

As we did for `Address`, `Action`, `Data` and other types (see #257), we can change the definition of `wallet.Sig` to `binary.Unmarshaler` and `binary.Marshaler`. Then

  1. The wallet backend can implement the `MarshalBinary`, `UnmarshalBinary` functions. that convert these types to/from binary format.
  2. These type can then be directly decoded using `perunio.(En|De)code` functions.

manoranjith:
Closes #292


backend/ethereum/channel/adjudicator.go
```diff
@@ -81,7 +83,11 @@ func (a *Adjudicator) Progress(ctx context.Context, req channel.ProgressReq) err
 		state adjudicator.ChannelState,
 		_ [][]byte,
 	) (*types.Transaction, error) {
-		return a.contract.Progress(opts, params, state, ethNewState, ethActorIndex, req.Sig)
+		sigBytes, err := req.Sig.MarshalBinary()
+		if err != nil {
+			return nil, errors.WithMessage(err, "converting to byte array")
+		}
+		return a.contract.Progress(opts, params, state, ethNewState, ethActorIndex, sigBytes)
```
matthiasgeihs, line 90: This looks a bit like you are mixing concerns here. MarshalBinary should typically be used by a marshaler that doesn't know much about the concrete type and just wants **some** byte representation. However, here we know the type and the contract expects a very specific byte representation.
```suggestion
		sig, ok := req.Sig.(*ethwallet.Sig)
		if !ok {
			return nil, errors.Errorf("invalid signature type: %T", req.Sig)
		}
		return a.contract.Progress(opts, params, state, ethNewState, ethActorIndex, *sig)
```

backend/ethereum/channel/adjudicator.go
```diff
@@ -105,12 +111,35 @@ func toEthSignedStates(subChannels []channel.SignedState) (ethSubChannels []adju
...
 	}
 	return
 }
 
+func sigsToByteArrays(signs []wallet.Sig) [][]byte {
+	byteArrays := make([][]byte, len(signs))
+	var err error
+	for i := range signs {
+		if signs[i] != nil {
+			byteArrays[i], err = signs[i].MarshalBinary()
```
matthiasgeihs, line 125: Same comments as above (regarding concerns of `MarshalBinary` vs byte representation for a contract call) applies here.

backend/ethereum/channel/withdraw.go
```diff
@@ -168,7 +168,12 @@ func (a *Adjudicator) newWithdrawalAuth(request channel.AdjudicatorReq, asset as
 	}
 
 	sig, err := request.Acc.SignData(enc)
-	return auth, sig, errors.WithMessage(err, "sign data")
+	if err != nil {
+		return assetholder.AssetHolderWithdrawalAuth{}, nil, errors.WithMessage(err, "sign data")
+	}
+	sigBytes, err := sig.MarshalBinary()
```
matthiasgeihs, line 174: same comments as above regarding byte representation for contract call applies here.

backend/ethereum/wallet/backend.go
```diff
@@ -46,31 +34,35 @@ func (b *Backend) NewAddress() wallet.Address {
...
 	hash := PrefixedHash(msg)
-	sigCopy := make([]byte, SigLen)
-	copy(sigCopy, sig)
-	if len(sigCopy) == SigLen && (sigCopy[SigLen-1] >= sigVSubtract) {
-		sigCopy[SigLen-1] -= sigVSubtract
+	ethSigCopy := make([]byte, sigLen)
+	copy(ethSigCopy, *ethSig)
+	if ethSigCopy[sigLen-1] >= sigVSubtract {
+		ethSigCopy[sigLen-1] -= sigVSubtract
 	}
```
matthiasgeihs, line 63: Should add a comment that explains what is happening here.

backend/ethereum/wallet/hd/account.go
```diff
@@ -35,14 +35,15 @@ func (a *Account) Address() wallet.Address {
...
 
 // SignData is used to sign data with this account.
-func (a *Account) SignData(data []byte) ([]byte, error) {
+func (a *Account) SignData(data []byte) (wallet.Sig, error) {
 	hash := crypto.Keccak256(data)
 	sig, err := a.wallet.SignText(a.Account, hash)
 	if err != nil {
 		return nil, errors.Wrap(err, "SignText")
 	}
 	sig[64] += 27
```
matthiasgeihs, line 44: Should add a comment that explains what is happening here.

Should use a const instead of `27`. Why is the linter not catching this?

backend/ethereum/wallet/keystore/account.go
```diff
@@ -34,14 +34,15 @@ func (a *Account) Address() wallet.Address {
...
 
 // SignData is used to sign data with this account.
-func (a *Account) SignData(data []byte) ([]byte, error) {
+func (a *Account) SignData(data []byte) (wallet.Sig, error) {
 	hash := ethwallet.PrefixedHash(data)
 	sig, err := a.wallet.Ks.SignHash(a.Account, hash)
 	if err != nil {
 		return nil, errors.Wrap(err, "SignHash")
 	}
 	sig[64] += 27
```
matthiasgeihs, line 43: Should add a comment that explains what is happening here.

Should use a const instead of `27`. Why is the linter not catching this?

(Could move this part to a central location that is used by the other Ethereum wallet implementations as well.)

backend/ethereum/wallet/sig.go
```diff
@@ -0,0 +1,57 @@
...
+	// ref https://godoc.org/github.com/ethereum/go-ethereum/crypto/secp256k1#Sign
+	// ref https://github.com/ethereum/go-ethereum/blob/54b271a86dd748f3b0bcebeaf678dc34e0d6177a/crypto/signature_cgo.go#L66
+	sigLen = 65
+
+	// sigVSubtract value that is subtracted from the last byte of a signature if
+	// the last bytes exceeds it.
+	sigVSubtract = 27
+)
+
+// Sig represents a signature generated using an ethereum account.
```
matthiasgeihs, line 34:
```suggestion
// Sig represents a signature generated using an Ethereum account.
```

backend/ethereum/wallet/sig.go
```diff
@@ -0,0 +1,57 @@
...
+package wallet
+
+import (
+	"fmt"
+
+	"perun.network/go-perun/wallet"
+)
+
+const (
+	// sigLen length of a signature in byte.
```
matthiasgeihs, line 24:
```suggestion
	// sigLen is the length of a signature in byte.
```

backend/ethereum/wallet/simple/account.go
```diff
@@ -37,14 +37,15 @@ func (a *Account) Address() wallet.Address {
...
 
 // SignData is used to sign data with this account.
-func (a *Account) SignData(data []byte) ([]byte, error) {
+func (a *Account) SignData(data []byte) (wallet.Sig, error) {
 	hash := ethwallet.PrefixedHash(data)
 	sig, err := a.SignHash(hash)
 	if err != nil {
 		return nil, errors.Wrap(err, "SignHash")
 	}
 	sig[64] += 27
```
matthiasgeihs, line 46: Should add a comment that explains what is happening here.

Should use a const instead of `27`. Why is the linter not catching this?

(Could move this part to a central location that is used by the other Ethereum wallet implementations as well.)

backend/sim/wallet/sig.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Sig represents the signature generated by the sim wallet.
+type Sig struct {
+	r, s *big.Int
+}
+
+// MarshalBinary marshals the signature into its binary representation. Error
+// will always be nil, it is for implementing BinaryMarshaler.
+func (sig Sig) MarshalBinary() ([]byte, error) {
+	pointSize := curve.Params().BitSize / bitsPerByte
```
matthiasgeihs, line 39: Why you redefine `pointSize` here?
```suggestion
```

backend/sim/wallet/sig.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"fmt"
+	"math/big"
+
+	"perun.network/go-perun/wallet"
+)
+
+var (
+	pointSize = curve.Params().BitSize / bitsPerByte
```
matthiasgeihs, line 25: Add comment?
```suggestion
	// pointSize is the size of a point on the curve in byte.
	pointSize = curve.Params().BitSize / bitsPerByte
```

backend/sim/wallet/wallet_internal_test.go
```diff
@@ -24,36 +24,19 @@ import (
...
+
 	pkgtest "polycry.pt/poly-go/test"
 )
 
-// TestSignatureSerialize tests serializeSignature and deserializeSignature since
-// a signature is only a []byte, we cant use io.serializer here.
-func TestSignatureSerialize(t *testing.T) {
-	a := assert.New(t)
-	// Constant seed for determinism
+func Test_Sig_GenericMarshaler(t *testing.T) {
```
matthiasgeihs, line 32: - Please try to adhere to function naming conventions. Underscore should be avoided.
https://ieftimov.com/post/testing-in-go-naming-conventions/#function-naming-conventions

- Please add function documentation.
```suggestion
// TestSignatureMarshalling tests marshalling and unmarshalling for type `Sig`.
func TestSignatureMarshalling(t *testing.T) {
```

channel/persistence/test/channel_internal_test.go
```diff
@@ -15,22 +15,33 @@
...
+	acc := test.NewRandomAccount(prng)
+	data1 := make([]byte, prng.Int63n(50))
+	prng.Read(data1)
+	data2 := make([]byte, prng.Int63n(50))
+	prng.Read(data2)
+
+	sig1, err := acc.SignData(data1)
+	require.NoError(t, err)
+	sig2, err := acc.SignData(data2)
+	require.NoError(t, err)
```
matthiasgeihs, line 37: Again, would be great if you could add a one-liner what is being tested here.

backend/sim/wallet/wallet_internal_test.go
matthiasgeihs: You could make this a black box test if you create a constructor for `Sig`.

channel/persistence/test/channel_internal_test.go
```diff
@@ -15,22 +15,33 @@
...
-	"polycry.pt/poly-go/test"
+	"perun.network/go-perun/wallet/test"
+	pkgtest "polycry.pt/poly-go/test"
 )
 
 func TestRequireEqualSigsTX(t *testing.T) {
-	prng := test.Prng(t)
+	prng := pkgtest.Prng(t)
+	acc := test.NewRandomAccount(prng)
+	data1 := make([]byte, prng.Int63n(50))
```
matthiasgeihs, line 29: Please use a const instead of `50`. The const name should describe the role of this value. (Wondering why the linter does not catch this.)

channel/persistence/test/channel_internal_test.go
```diff
@@ -15,22 +15,33 @@
...
+	prng.Read(data1)
+	data2 := make([]byte, prng.Int63n(50))
```
matthiasgeihs, line 31: Please use a const instead of `50`. The const name should describe the role of this value. (Wondering why the linter does not catch this.)

client/updatemsgs.go
```diff
@@ -142,8 +142,8 @@ func (c *msgChannelUpdate) Decode(r io.Reader) (err error) {
 	if err := perunio.Decode(r, c.State, &c.ActorIdx); err != nil {
 		return err
 	}
-	c.Sig, err = wallet.DecodeSig(r)
-	return err
+	c.Sig = wallet.NewSig()
+	return perunio.Decode(r, c.Sig)
```
matthiasgeihs, line 146: So there is still a direct dependency from `client` onto `perunio`?

client/updatemsgs.go
```diff
@@ -154,8 +154,8 @@ func (c *msgChannelUpdateAcc) Decode(r io.Reader) (err error) {
 	if err := perunio.Decode(r, &c.ChannelID, &c.Version); err != nil {
 		return err
 	}
-	c.Sig, err = wallet.DecodeSig(r)
-	return err
+	c.Sig = wallet.NewSig()
+	return perunio.Decode(r, c.Sig)
```
matthiasgeihs, line 158: So there is still a direct dependency from `client` onto `perunio`?

wallet/sig.go
```diff
@@ -24,8 +24,22 @@ import (
 	"perun.network/go-perun/wire/perunio"
 )
 
-// Sig is a single signature.
-type Sig = []byte
+// Sig represents a single signature generated over a given data using an
+// account. It is usually a byte array and the length depends upon the specific
+// signature scheme being used, which in turn depends upon the wallet backend.
+//
+// It is sent over wire as a binary data.
```
matthiasgeihs, line 31: This part is not useful to mention here as it explains the functionality of a different package.
```suggestion
//
// It is sent over wire as a binary data.
```

wallet/sig.go
```diff
@@ -99,7 +116,8 @@ func DecodeSparseSigs(r io.Reader, sigs *[]Sig) (err error) {
 			if ((mask[maskIdx] >> bitIdx) % binaryModulo) == 0 {
 				(*sigs)[sigIdx] = nil
 			} else {
-				(*sigs)[sigIdx], err = DecodeSig(r)
+				(*sigs)[sigIdx] = NewSig()
+				err = perunio.Decode(r, (*sigs)[sigIdx])
```
matthiasgeihs, line 120: So there is still a direct dependency from `wallet` onto `perunio`?

wallet/test/wallet.go
```diff
@@ -111,15 +118,20 @@ func GenericSignatureSizeTest(t *testing.T, s *Setup) {
 	sign, err := acc.SignData(s.DataToSign)
 	require.NoError(t, err, "Sign with unlocked account should succeed")
 
+	signRaw, err := sign.MarshalBinary()
```
matthiasgeihs, line 121:
```suggestion
	sigRaw, err := sig.MarshalBinary()
```

wallet/test/wallet.go
```diff
@@ -89,18 +85,29 @@ func TestAccountWithWalletAndBackend(t *testing.T, s *Setup) { //nolint:revive /
...
+
+	sigBytes, err := sig.MarshalBinary()
+	require.NoError(t, err)
+	tamperer(&sigBytes)
+	return sigBytes
+}
+
 // GenericSignatureSizeTest tests that the size of the signatures produced by
 // Account.Sign(…) does not vary between executions (tested with 2048 samples).
 func GenericSignatureSizeTest(t *testing.T, s *Setup) {
```
matthiasgeihs, line 113: I think we can throw that test out. Where do we specify that a signature byte representation must always have the same length?

matthiasgeihs:

PR #295, opened 2021-12-14
Move verify function from wallet.Backend to wallet.Sig

### LINKED ISSUE #294: Move VerifySignature from wallet.Backend to wallet.Sig
### Location

wallet, ethereum/wallet, sim/wallet

### Problem

Previously, VerifySignature function is defined on the wallet backend. This looked fine, because until recently, wallet.Sig was defined as a byte array. But, now it has been redefined as an interface with each wallet implementation implementing the wallet.Sig interface (see #292). Now, it looks odd to use the `wallet.Backend.VerifySignature` method when `Sig` itself is defined as a separate type.

### Proposal

Move the VerifySignature method from `wallet.Backend` to `wallet.Sig` interface.

manoranjith:
Closes #294.


matthiasgeihs:
This is a change to the external API with no clear justification. It is debatable which is the better approach, have the Verify method on the Wallet, or on the Sig.

I have looked at https://pkg.go.dev/crypto/ecdsa and [https://github.com/google/tink](https://github.com/google/tink/blob/master/go/signature/signature_test.go) and both do not have the Verify function on the signature. Instead, they rather have a signature of type `[]byte` and have the Verify function on some central object, either the package or a dedicated Verifier object.

The fact that other libraries tend to treat signatures typically just as `[]byte` makes me wonder whether actually introducing a type `Sig` is a good idea at all. It seems to introduce a lot of additional types (e.g., now each wallet needs to implement a type `Sig`). Moreover, you mentioned that the reason for introducing type `Sig` was separation of concerns for `perunio`, but now I am wondering: If it is about serializing a signature into bytes, shouldn't that part be trivial if the type of sig is `[]byte` already?

matthiasgeihs (commented):
Please see https://github.com/hyperledger-labs/go-perun/pull/295#issuecomment-993654475

matthiasgeihs:

PR #300, opened 2022-01-19
Fix virtual channel proposal id generation

### LINKED ISSUE #299: Correctly generate the ID for virtual channel proposal
### Location

[client]

### Problem

The ID for ledger channel and sub channel proposals, is generated by hashing all the parameters.

But, the ID for virtual channel proposal, is generated only using the base channel proposal parameters. Because, unlike, `LedgerChannelProposal` and `SubChannelProposal` types, the `VirtualChannelProposal` types does not have a `ProposalID` method defined for it. Hence, it inherits the `ProposalID` method of the `BaseChannelProposal`. See [here](https://github.com/hyperledger-labs/go-perun/blob/4e91b2a559e34040d933e2f3b7610964ffa545dd/client/proposalmsgs.go#L196).

### Proposal

Remove the `ProposalID` method on the base channel proposal. And add a `ProposalID` method on the `VirtualChannelProposal`, which generates the ID using all the parameters of a virtual channel proposal.

manoranjith:
- Previously, the proposal ID for a virtual channel was generated using
  the ProposalID method defined on BaseChannelProposal.

- Hence, not all parameters of the virtual channel proposal were used.

- Fix it by defining a ProposalID method on the virtual channel.

- Also, remove the ProposalID method on the BaseChannelProposal, as it
  is not needed.

Fixes #299.

### DIFF
--- a/client/proposalmsgs.go
+++ b/client/proposalmsgs.go
@@ -192,17 +192,6 @@ func (p BaseChannelProposal) Encode(w io.Writer) error {
 		optAppAndDataEnc, p.InitBals, p.FundingAgreement)
 }
 
-// ProposalID returns the identifier of this channel proposal.
-func (p BaseChannelProposal) ProposalID() (propID ProposalID) {
-	hasher := newHasher()
-	if err := perunio.Encode(hasher, p.Base()); err != nil {
-		log.Panicf("proposal ID base encoding: %v", err)
-	}
-
-	copy(propID[:], hasher.Sum(nil))
-	return
-}
-
 // Decode decodes a BaseChannelProposal from an io.Reader.
 func (p *BaseChannelProposal) Decode(r io.Reader) (err error) {
 	if p.InitBals == nil {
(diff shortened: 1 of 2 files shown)


client/proposalmsgs.go
```diff
@@ -620,6 +609,33 @@ func (VirtualChannelProposal) Type() wire.Type {
...
+		for j := range p.IndexMaps {
+			if err := perunio.Encode(hasher, (p.IndexMaps[i][j])); err != nil {
+				log.Panicf("proposal ID base encoding: %v", err)
+			}
+		}
+	}
+
+	copy(propID[:], hasher.Sum(nil))
+	return
+}
```
matthiasgeihs, line 637: We can reuse the `VirtualChannelProposal.Encode` method:
```suggestion
// ProposalID returns the identifier of this channel proposal.
func (p VirtualChannelProposal) ProposalID() (propID ProposalID) {
	hasher := newHasher()
	if err := perunio.Encode(hasher, p); err != nil {
		log.Panicf("encoding proposal: %v", err)
	}
	copy(propID[:], hasher.Sum(nil))
	return
}
```

matthiasgeihs (changes requested):
I think we can achieve the same thing by using `VirtualChannelProposal.Encode`.

client/proposalmsgs.go
manoranjith: Yeah, I agree.

manoranjith: Updated in f3c6404. Did the same for ledger channel and sub channel proposals, in the following commit: 4576ba2.

manoranjith: As it was the only suggestion, I squashed the review commit and rebased the branch onto develop.

matthiasgeihs (approved):

PR #304, opened 2022-01-21
303 check assets

### LINKED ISSUE #303: validTransition does not check assets
### Location
[machine. validTransition](https://github.com/hyperledger-labs/go-perun/blob/d960e1c6cff4e53990f5e54bf5c4ea5e799049e8/channel/machine.go#L486)

### Problem

In [machine. validTransition](https://github.com/hyperledger-labs/go-perun/blob/d960e1c6cff4e53990f5e54bf5c4ea5e799049e8/channel/machine.go#L486), we check whether the total balances remain constant during a state transition. However, we do not check whether the assets also remain the same, which makes the balances check non-sensical.

### Proposal

Since we intended to check that the total assets remain constant, we also check `alloc.Assets` does not change. In the long run, we might also move `Assets` to the channel parameters, unless we find an application that requires the assets to change.

matthiasgeihs:
Closes #303

### DIFF
--- a/channel/allocation.go
+++ b/channel/allocation.go
@@ -582,7 +582,7 @@ func (a *Allocation) Equal(b *Allocation) error {
 		return nil
 	}
 	// Compare Assets
-	if err := AssetsAssertEqual(a.Assets, b.Assets); err != nil {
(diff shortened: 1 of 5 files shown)


sebastianst (approved):
Beautiful ✨

PR #305, opened 2022-01-24
Rename wire message types

### LINKED ISSUE #302: Off-chain messages: Naming conventions and type export for wire serialization
### Location

[client, wire]

### Problem

Currently, multiple conventions are used for naming structs that represent off-chain messages.

1. Exported types and suffix `Msg` (see wire/controlmsgs.go): ShutdownMsg, PingMsg etc.,
2.  Exported types with no suffix or prefix (see client/proposalmsgs.go): LedgerChannelProposal, LedgerChannelProposalAcc etc.,
3. Unexported types with no suffix or prefix (see client/updatemsgs.go): virtualChannelFundingProposal and virtualChannelSettlementProposal
4. Unexported types with prefix `msg` (see client/updatemsgs.go): msgChannelUpdate, msgChannelUpdateAcc

The has two issues:

1. For implementing the protobuff adapter, all the types that represent off-chain messages must be exported.
2. It would be better to use a uniform naming convention, to make it more readable.

### Proposal

My proposal is to consistently use the first style (exported types with suffix `Msg` for all the types.

manoranjith:
- Consistently use the pattern `<message-name>Msg` for all message types.

- Export the unexported message types, so that newer wire serialization adapters can define functions for serializing/de-serializing these messages.

Closes #302

### DIFF
--- a/client/channel.go
+++ b/client/channel.go
@@ -209,7 +209,7 @@ func (c *Channel) initExchangeSigsAndEnable(ctx context.Context) error {
 
 	send := make(chan error)
 	go func() {
-		send <- c.conn.Send(ctx, &msgChannelUpdateAcc{
+		send <- c.conn.Send(ctx, &ChannelUpdateAccMsg{
 			ChannelID: c.ID(),
 			Version:   0,
 			Sig:       sig,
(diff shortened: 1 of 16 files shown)


matthiasgeihs (commented):
OK, after seeing the changes applied, I realize something that I didn't when reviewing the corresponding issue: This will change the name of types that we actually use externally (like LedgerChannelProposal), plus we might want to change the type name again once we move them to a `message` package (in which case we can drop the suffix).

I see the following alternatives:
- Option 2: Only change the name of the previously unexported types because we need to change these anyways. This will ensure compatibility with existing implementations. And later we can make the naming scheme consistent for a longer period of time, once we moved the message types to a more final location.
- Option 3: Move all message types to a separate package `client.message` and drop the suffix.

What do you think?

manoranjith:
Moving messages is already described in step 7 of #233. Because, it is a more involving task, we had though if doing it sometime later.

So, between 2 and 3, I would prefer 2.

matthiasgeihs (approved):
OK, after discussion, we decided to keep as proposed in the issue description.

PR #306, opened 2022-01-25
:boom: Channel.Update/ForceUpdate: updater no longer returns error

### LINKED ISSUE #155: Redesign channel updates API
Currently, the go-perun client exposes two APIs for updating a channel: [`Update`](https://github.com/hyperledger-labs/go-perun/blob/749b197450cf4e82e302e5ff2552a3ee744e265e/client/update.go#L140) and [`UpdateBy'](https://github.com/hyperledger-labs/go-perun/blob/749b197450cf4e82e302e5ff2552a3ee744e265e/client/update.go#L236).

```go
func (c *Channel) Update(ctx context.Context, next *channel.State) (err error) {...}
func (c *Channel) UpdateBy(ctx context.Context, update func(*channel.State) error) (err error) {...}
```

In both cases, the user can modify any property of the state:
1. Version
2. Any field in Allocation (Assets, Balances or Locked funds that contain sub-allocations).
3. Data
4. IsFinal flag

In many scenarios, arbitrarily modifying these fields would lead to invalid updates that will fail. For instance,
1. Version must be incremented by 1 for each update, but this API allows the user to arbitrarily modify the version.
2. Assets field in Allocation must not be changed. It should only read by the user.
3. Locked funds should not be modified by the user. It is meant only for those updates that fund/withdraw sub/virtual channels. These updates will be done only by the SDK and not by the user.
4. IsFinal flag can be set only when certain conditions are met. Eg: For ledger channels, there should be no locked funds.

It will a lot of overhead on the user to ensure these conditions when making an update.
Hence, I propose to redesign the update API to make it more easy to use. It can replace both the existing APIs:

```go
Update(updater func(Balances, Data) error, finalize bool) (err error) {...}
```
In this API, 
1. User is able to modify only the balances and data. 
    1. API will validate if sum of balances remain constant and if data is valid for the given app.
2. If user needs info on Assets or App, they could read it from the channel. These are set during channel opening and never change during the lifetime of the channel.
3. Finalize flag can be used to finalize a channel. The API will apply the finalize flag only after ensuring if necessary conditions are met.

matthiasgeihs:
Related to #155 

Instead of changing the API drastically, improve documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ func main() {
 	}()
 
 	// send a channel update request to the other channel peer(s)
-	err = ch.Update(ctx, func(s *channel.State) error {
+	err = ch.Update(ctx, func(s *channel.State) {
 		// update state s, e.g., moving funds or changing app data
 	})
 	if err != nil { /* handle error */ }
--- a/client/adjudicate.go
+++ b/client/adjudicate.go
@@ -210,6 +210,11 @@ func (c *Channel) ForceUpdate(ctx context.Context, updater func(*channel.State))
 	updater(state)
 	state.Version++
 
+	// Check state transition.
+	if err := c.machine.ValidTransition(state); err != nil {
+		return errors.WithMessage(err, "validating state transition")
+	}
+
 	// Apply state in machine and generate signature
 	if err := c.machine.SetProgressing(ctx, state); err != nil {
 		return errors.WithMessage(err, "updating machine")
--- a/client/adjudicate.go
+++ b/client/adjudicate.go
@@ -178,16 +178,19 @@ func (c *Channel) registerDispute(ctx context.Context) error {
 	return nil
 }
 
-// ForceUpdate enforces a channel update on the adjudicator. `update` gets as
-// input a copy of the current state and is expected to modify the state as
-// intended. It should not to increment the version number as this is
-// automatically.
+// ForceUpdate enforces a state update through the adjudicator as specified by
+// the `updater` function.
+//
+// The updater function must adhere to the following rules:
+// - The version must not be changed. It is incremented automatically.
+// - The assets must not be changed.
+// - The locked allocation must not be changed.
 //
 // Returns TxTimedoutError when the program times out waiting for a transaction
 // to be mined. Returns ChainNotReachableError if the connection to the
 // blockchain network fails when sending a transaction to / reading from the
 // blockchain.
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -146,7 +146,7 @@ func main() {
 	}()
 
 	// send a channel update request to the other channel peer(s)
-	err = ch.Update(ctx, func(s *channel.State) error {
+	err = ch.Update(ctx, func(s *channel.State)) error {
```
manoranjith, line 149: With the updated API, this should be `err = ch.Update(ctx, func(s *channel.State) {`

client/test/channel.go
```diff
@@ -98,7 +98,7 @@ func (ch *paymentChannel) acceptSubchannel(
 	return subchannel
 }
 
-func (ch *paymentChannel) sendUpdate(update func(*channel.State) error, desc string) {
+func (ch *paymentChannel) sendUpdate(update func(*channel.State), desc string) {
```
manoranjith, line 101:
```suggestion
func (ch *paymentChannel) sendUpdate(updater func(*channel.State), desc string) {
```

`updater` looks more appropriate and consistent with the previous change.

client/adjudicate.go
```diff
@@ -204,11 +207,8 @@ func (c *Channel) ForceUpdate(ctx context.Context, update func(*channel.State) e
 
 	// Update state
 	state := ar.Tx.State.Clone()
+	updater(state)
```
manoranjith, line 210: I notice that, unlike the `channel.Update` function, we do not validate the state after applying the `updater`.

Is it not necessary to validate the state, before sending an on-chain transaction ? Because, if the `updater` has modified the state in unexpected ways (say, decrementing version number), then this would mean sending a transaction, that would fail for sure.

manoranjith (changes requested):
Have a requested a change in the Readme file and added two more questions.

client/adjudicate.go
matthiasgeihs: You are right. Added a quick fix for that.

matthiasgeihs:
@manoranjith Addressed all comments, please have another look.

client/adjudicate.go
manoranjith: The fix looks good to me.

manoranjith:
@matthiasgeihs The fix looks good to me. But there's a failing test, could you have a look into it ? Also, you may want to add a `Signed-off-by:...` line in the last commit.

manoranjith (approved):

PR #307, opened 2022-01-27
Update the mechanism for propsal ID generation

### LINKED ISSUE #301: Mechanism for proposal ID generation
### Location

[client]

### Problem

Currently, we use the following mechanism for generating the proposal ID:
  1. Encode all the parameters of the proposal, using perunio encoding into a binary format.
  2. Hash the encoded data.

In #233, we have abstracted out the wire encoding component, so that different encoding formats can be used.
We also need to update the mechanism for generating proposal IDs, so that it is not dependent on a particular wire encoding format.  

### Proposal

There are two possible options:

1. We could add an interface in the wire package, that exposes methods for encoding proposals. 
These functions, will in turn use the wire serialization format that is set by the user. This can be perunio or any other protocol.
```
type ProposalEncoder {
    LedgerChannelProposal(p LedgerChannelProposal) ([]byte, error)
    SubChannelProposal(p SubChannelProposal) ([]byte, error)
    VirtualChannelProposal(p VirtualChannelProposal) ([]byte, error)
}
```

*Implications of this approach*:

In case of protocol buffers (the second wire serialization protocol, that we want to support), the serialized output is not canonical across different languages. Hence, serializing the proposal, could result in different binary outputs, when using  implementations of protocol buffer library in different languages: Say, c and golang. 
References: [1](https://gist.github.com/kchristidis/39c8b310fd9da43d515c4394c3cd9510), [2](https://github.com/golang/protobuf/issues/1121#issuecomment-627554847) 

This would mean that, perun clients implemented in different languages, might generate different proposal IDs for the same proposal.

2. Looking at the fields in the channel proposals ([here](https://github.com/hyperledger-labs/go-perun/blob/aafb18b97be5a32c76215a3965dab6f3556c6900/client/proposalmsgs.go#L106)), most of the fields are already byte arrays. Apart from that, we have only one field of `uint64` type and few fields of `big.Int` type.

    a. For the byte array types, we could directly hash them.
    b. For uint64, we could convert them to their binary representation (big/little endian), which is easy to do, consistently across programming languages.
    c. For big.Int or multi-precision integers, we could convert them to their decimal representation, strip the leading zeros and hash the resulting string.
    
*Implications of this approach*:

As we directly use the byte arrays and define custom canonical formats for the other types (uint64, multi-precision integer), this mechanism will generate the same ID for a given proposal, across different programming languages. However, this introduces a small overhead of having to implement this across different implementations and maintain them.

manoranjith:
As described in #301, use the proposer nonce share as the proposal ID.

Also,

- Update the tests for proposal ID generation by removing the assertions that
  check if the proposal ID changes when the channel parameters change.

- Make the ProposalID a field of BaseChannel proposal, as it does not depend on
  the type of channel proposal and remove the proposal ID method from
  `ChannelProposal` interface.

Closes #301.


matthiasgeihs:
Seb mentioned that this will not work if the client uses incrementing nonces per participant set (he could do that for whatever reason). So two ideas: 1) Either we generate the ProposalID separately. 2) We remove the option to set a nonce manually and always draw it uniform random.

What do you think? @manoranjith

manoranjith:
I think, we can leave the nonce part as such and generate proposal id separately, using a uniform random source.

matthiasgeihs (changes requested):
decided to change implementation according to comments above

manoranjith:
Have updated the PR. Now the proposal ID is generated using a random source, independent of the nonce share.
Can you have a look @matthiasgeihs. 

If it looks good, then I can squash the commit and rebase onto main.

matthiasgeihs (dismissed):
LGTM

matthiasgeihs:
@manoranjith Can you rebase?

manoranjith:
I have rebased. I am currently, extracting the points from our discussions and writing a proposal that describes the mechanism for proposal ID and the rationale behind it.

Do you think, we should wait for it to be completed and include its reference in the commit message before merging ?

matthiasgeihs:
For me it's fine since we agree about the changes. In this sense I see the proposal as optional here. I would propose to merge if you agree.

matthiasgeihs (approved):

PR #309, opened 2022-01-31
eth/channel: Fix tx receipt pulling

### LINKED ISSUE #308: ContractBackend.confirmNTimes fails with "pulling receipt"
### Location

https://github.com/hyperledger-labs/go-perun/blob/f3f8a24b496695824e821568825758cfd4131bab/backend/ethereum/channel/contractbackend.go#L203

### Problem
`confirmNTimes` pulls the transaction receipt at least two times. Once for confirming that the transaction has been included in the chain, and then each time a new block is mined (until the transaction is confirmed sufficiently often). We observed in the context of the web app (Goerli Network) that it can happen that the tx is first found, and then not found anymore. In this case, `confirmNTimes` fails with "pulling receipt".

### Proposal
Instead of failing, the logic should instead continue to wait for the transaction to be confirmed. In fact, such an error could have happened due to a reorg, which means that we probably would want to reset the transaction confirmation counter in such a case. (I am currently unsure if we do this already.)

matthiasgeihs:
Closes #308

### DIFF
--- a/backend/ethereum/channel/contractbackend.go
+++ b/backend/ethereum/channel/contractbackend.go
@@ -227,7 +227,8 @@ func (c *ContractBackend) confirmNTimes(ctx context.Context, tx *types.Transacti
 			receipt, err := c.ContractInterface.TransactionReceipt(ctx, tx.Hash())
 			if err != nil {
 				err = cherrors.CheckIsChainNotReachableError(err)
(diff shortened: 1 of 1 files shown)


manoranjith (approved):
LGTM.

manoranjith:
@matthiasgeihs Could you rebase the branch onto tip of main ?

manoranjith (approved):

PR #313, opened 2022-02-04
client: Export proposer and proposee index

### LINKED ISSUE #79: Indicate proposer index in channel proposal

matthiasgeihs:
Closes #79

### DIFF
--- a/client/proposal.go
+++ b/client/proposal.go
@@ -30,7 +30,12 @@ import (
 	"polycry.pt/poly-go/sync/atomic"
 )
 
-const proposerIdx, proposeeIdx = 0, 1
+const (
+	// ProposerIdx is the index of the channel proposer.
+	ProposerIdx = 0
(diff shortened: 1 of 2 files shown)


RmbRT (approved):
I think when we some day switch to 3+-party channels, then we will have to rethink our modelling of this, but for now, it's fine.

PR #315, opened 2022-02-04
channel: Init balances to zero

matthiasgeihs:
Fixes a bug where balances were not initialized to zero.


RmbRT (approved):
This is one of the cases where it's sub-optimal that `nil` and `0` values have to be handled differently. Although for many cases, this change creates a double initialisation of values, it does make the code more robust. In another language, wrapper classes and default constructors would have spared us a lot of trouble here.

RmbRT (approved):

PR #316, opened 2022-02-08
Add missing tests for sync msg serialization

manoranjith:
- These tests were missing.

### DIFF
--- a/client/syncmsgs_test.go
+++ b/client/syncmsgs_test.go
@@ -1,0 +1,40 @@
+// Copyright 2022 - See NOTICE file for copyright holders.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


client/syncmsgs_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"testing"
+
+	"perun.network/go-perun/channel"
+	"perun.network/go-perun/channel/test"
+	"perun.network/go-perun/client"
+	peruniotest "perun.network/go-perun/wire/perunio/test"
+	pkgtest "polycry.pt/poly-go/test"
+)
+
+func TestChannelSyncSerialization(t *testing.T) {
```
matthiasgeihs, line 27: You could add a short comment what this test does.

matthiasgeihs (dismissed):

client/syncmsgs_test.go
manoranjith: I think the name of the test already explains that this tests the serialization of channel sync message. I have followed the same naming conventions, as that is used for other Serialization tests in `proposamsgs_test.go` and `updatemsgs_test.go`.

matthiasgeihs (approved):

PR #318, opened 2022-02-08
Implement protobuf adapter for wire serialization

### LINKED ISSUE #311: Add protobuf wire encoding
We recently introduced an abstraction that makes it possible to inject a custom wire encoding. In particular, this was done for enabling protobuf based encoding. The next step is to write such an encoder.

- [x] #316 
- [x] #317 
- [ ] #318

manoranjith:
Closes #311


client/proposalmsgs_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	"perun.network/go-perun/channel/test"
 	"perun.network/go-perun/client"
 	clienttest "perun.network/go-perun/client/test"
+	_ "perun.network/go-perun/wire/perunio/serializer" // wire serialzer init
```
matthiasgeihs, line 27: Could we move this file to `wire/perunio/test` since it seems to be specific to perunio?

wire/controlmsgs_test.go
```diff
@@ -17,6 +17,7 @@ package wire_test
 import (
 	"testing"
 
+	_ "perun.network/go-perun/wire/perunio/serializer" // wire serialzer init
```
matthiasgeihs, line 20: Could we move this file to wire/perunio/test since it seems to be specific to perunio?

wire/protobuf/serializer.go
```diff
@@ -0,0 +1,145 @@
...
+	"perun.network/go-perun/wire"
+)
+
+// Serializer implements methods for encoding/decoding
+// envelopes using protobuf encoding.
+type Serializer struct{}
+
+// func init() {
+// 	wire.SetEnvelopeSerializer(Serializer{})
+// }
```
matthiasgeihs, line 33: Hm?

wire/protobuf/serializer.go
```diff
@@ -0,0 +1,145 @@
...
+	if err := env.Sender.UnmarshalBinary(protoEnv.Sender); err != nil {
+		return nil, errors.Wrap(err, "unmarshalling sender address")
+	}
+	env.Recipient = wire.NewAddress()
+	if err := env.Recipient.UnmarshalBinary(protoEnv.Recipient); err != nil {
+		return nil, errors.Wrap(err, "unmarshalling recipient address")
+	}
+
+	switch protoEnv.Msg.(type) {
+	case *Envelope_PingMsg:
```
matthiasgeihs, line 126: I realize that this PR is not complete yet as many message types are missing, right?

wire/test/msgstest.go
```diff
@@ -46,8 +46,12 @@ func AuthMsgsTest(t *testing.T, serializerTest func(t *testing.T, msg wire.Msg))
 
 // newRandomstring returns a random ascii string of length between minLen and
 // minLen+maxLenDiff.
-func newRandomString(rng *rand.Rand, minLen, maxLenDiff int) string {
+func newRandomASCIIString(rng *rand.Rand, minLen, maxLenDiff int) string {
```
matthiasgeihs, line 49: You forgot to rename in the function comment.

Could it be a problem if protobuf only supports ASCII?

matthiasgeihs (commented):
Realized that my review came to early. Seems like not all messages are covered yet.

wire/protobuf/serializer.go
manoranjith: The PR is not yet complete. I will adding a few more commits, that will eventually remove this.

wire/protobuf/serializer.go
manoranjith: Yes.

wire/test/msgstest.go
manoranjith: I will rename in the comment.

The reason behind this change is that, protobuf encodes strings using UTF-8 encoding and hence supports all unicode characters. However, not all byte arrays are valid unicode strings (see [here](https://en.wikipedia.org/wiki/UTF-8#Encoding)) and protobuf errors when an invalid unicode string is passed to it.

Generating a random unicode string is a bit tricky, so I have used random ASCII string for tests.

manoranjith: With perunio encoding, this was not a problem because `io.WriteString`, by default, treats the string as byte array.

wire/net/test/connhub_internal_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	_ "perun.network/go-perun/backend/sim" // backend init
 	wallettest "perun.network/go-perun/wallet/test"
 	"perun.network/go-perun/wire"
+	_ "perun.network/go-perun/wire/protobuf" // wire serialzer init
```
matthiasgeihs, line 27: I don't like that the user now has to do these blank imports every time it uses perun. It's easy to forget. Is there a reminder in case you forget to set the serializer like this? What will happen then?

wire/protobuf/generate.go
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+package protobuf
+
+// This command generates the bindings for serializing wire messages using
+// protocol buffers serialization protocol.
+//
+// It requires protoc and protoc-gen-go to be installed.
+
+//go:generate protoc --go_out=. wire.proto
```
matthiasgeihs, line 22: Do we document this in the readme?

wire/protobuf/serializer.go
```diff
@@ -0,0 +1,182 @@
...
+	case *client.VirtualChannelFundingProposalMsg:
+		protoEnv.Msg, err = fromVirtualChannelFundingProposalMsg(msg)
+	case *client.VirtualChannelSettlementProposalMsg:
+		protoEnv.Msg, err = fromVirtualChannelSettlementProposalMsg(msg)
+	case *client.ChannelUpdateAccMsg:
+		protoEnv.Msg = fromChannelUpdateAccMsg(msg)
+	case *client.ChannelUpdateRejMsg:
+		protoEnv.Msg = fromChannelUpdateRejMsg(msg)
+	case *client.ChannelSyncMsg:
+		protoEnv.Msg, err = fromChannelSyncMsg(msg)
```
matthiasgeihs, line 72: Please throw an error in the default case.

wire/protobuf/serializer.go
```diff
@@ -0,0 +1,182 @@
...
+	case *Envelope_VirtualChannelFundingProposalMsg:
+		env.Msg, err = toVirtualChannelFundingProposalMsg(protoMsg)
+	case *Envelope_VirtualChannelSettlementProposalMsg:
+		env.Msg, err = toVirtualChannelSettlementProposalMsg(protoMsg)
+	case *Envelope_ChannelUpdateAccMsg:
+		env.Msg = toChannelUpdateAccMsg(protoMsg)
+	case *Envelope_ChannelUpdateRejMsg:
+		env.Msg = toChannelUpdateRejMsg(protoMsg)
+	case *Envelope_ChannelSyncMsg:
+		env.Msg, err = toChannelSyncMsg(protoMsg)
```
matthiasgeihs, line 155: Please throw an error in the default case.

wire/test/msgstest.go
matthiasgeihs: What i meant is that now the behavior of some messages is different for different encodings. Probably not a problem in most cases, but in an edge case somebody might try to encode some information into this message as bytes which will work with the perunio encoding, but not with the protobuf encoding.

matthiasgeihs (commented):
A few comments.

wire/test/msgstest.go
manoranjith: Okay, now I get your point. I can see two possible solutions:

1. If these fields can hold arbitrary binary data, then we can change its type in protobuf adapter from string to byte array. Then, this restriction does not apply.
2. If these fields can hold only string data, then perunio adapter could probably validate the string, before encoding.

What do you think @matthias ?

wire/net/test/connhub_internal_test.go
manoranjith: I agree that this could be missed, in that case there will be a panic.

I think this (blank imports) is the way, we expect the user to initialize different backends (wallet, channel etc.,). So extended the same paradigm here.

wire/protobuf/generate.go
matthiasgeihs: No need to do this in the general README. We also don't do it for the contract bindings.

wire/test/msgstest.go
matthiasgeihs: OK, maybe we can add a comment to the type that this is expected to be a unicode string.

wire/controlmsgs_test.go
manoranjith: Yes, it is specific to `perunio`. But, the tests are in wire package because the Encode/Decode functions are implemented in this package.

wire/test/msgstest.go
manoranjith: Added the comment. See these two commits: 57e48716 and f08fb808.

wire/protobuf/serializer.go
manoranjith: Add in this commit: d995e181.

wire/protobuf/serializer.go
manoranjith: Add in this commit: d995e181.

manoranjith:
@matthiasgeihs  I have addressed all the review comments. I have also rebased the branch and squashed the review commits.

matthiasgeihs (approved):
👍

PR #322, opened 2022-02-09
Use LatestSigner and don't set GasPrice in ContractBackend

sebastianst:
- eth/channel: Use types.LatestSigner for all sim backend tests
  - Before, `types.EIP155Signer`s were created in many different places.
It is recommended by the geth folks to use `types.LatestSigner` instead.
This was streamlined an there's now a common `channel/test.SimSigner` that
can always be used together with the simulated backend.
  - `FundAddress` and `Test_ConfirmTransaction` were adapted to use the
new `types.DynamicFeeTx`.
  - The unused `test.GasLimit` variable was deleted.
  - The `test.GasPrice` variable was renamed to `InitialGasBaseFee` and its
description updated.
- eth/channel: Don't set `GasPrice` in `ContractBackend.NewTransactor`
  - Setting `TransactOpts.GasPrice` forces the transactions to be sent as a
legacy transaction, instead of a type 2 EIP1559, which results in much
higher gas costs.
  - Also remove the `SuggestedGasPrice` method from the `SimulatedBackend`, as
it is not used anymore, because EIP1559 TXs are now sent everywhere.

Opened as a draft to let @ndzik try it out first if we truly now send EIP1559 txs.
Original issue in Erdstall is at https://github.com/perun-network/erdstall-ext/issues/121

### DIFF
--- a/backend/ethereum/channel/test/simulated.go
+++ b/backend/ethereum/channel/test/simulated.go
@@ -106,11 +106,6 @@ func NewSimulatedBackend(opts ...SimBackendOpt) *SimulatedBackend {
 	return sb
 }
 
-// SuggestGasPrice always returns `GasPrice`.
-func (*SimulatedBackend) SuggestGasPrice(context.Context) (*big.Int, error) {
-	return big.NewInt(GasPrice), nil
-}
-
 // SendTransaction executes a transaction.
 func (s *SimulatedBackend) SendTransaction(ctx context.Context, tx *types.Transaction) error {
 	if err := s.SimulatedBackend.SendTransaction(ctx, tx); err != nil {
--- a/backend/ethereum/channel/test/simulated.go
+++ b/backend/ethereum/channel/test/simulated.go
@@ -37,15 +37,20 @@ import (
 )
 
 const (
-	// GasLimit is the max amount of gas we want to send per transaction.
-	GasLimit = 500000
-	// GasPrice is the gas price that is used for simulated transactions.
-	// This value is set to `maxFeePerGas` from go-ethereum to prevent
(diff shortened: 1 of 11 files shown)


backend/ethereum/channel/test/simulated.go
```diff
@@ -122,17 +122,22 @@ func (s *SimulatedBackend) FundAddress(ctx context.Context, addr common.Address)
 	if err != nil {
 		panic(err)
 	}
-	tx := types.NewTransaction(nonce, addr, test.MaxBalance, GasLimit, big.NewInt(GasPrice), nil)
-	signer := types.NewEIP155Signer(params.AllEthashProtocolChanges.ChainID)
-	signedTX, err := types.SignTx(tx, signer, s.faucetKey)
+	txdata := &types.DynamicFeeTx{
```
matthiasgeihs, line 125: Why use the concrete type `DynamicFeeTx`? Wouldn't `types.NewTransaction` automatically give the correct type?

sebastianst: `NewTransaction` has the comment
```
// NewTransaction creates an unsigned legacy transaction.
// Deprecated: use NewTx instead.
```
And `NewTx` takes any of `LegacyTx`, `DynamicFeeTx` or `AccessListTx` (no idea what the 3. is).

Note that I use `types.SignNewTx` below, which is just a shortcut for first calling `NewTx` followed by `SignTx`.

sebastianst: It made me realize that we still use `types.NewTransaction` in `test/transactor.go`. Will also change them to the new tx types.

PR #325, opened 2022-02-16
Initialize wire encoding explicitly

matthiasgeihs:
This PR proposes to initialize the wire encoding explicitly when needed, instead of setting it via an implicit import. This allows a user to have a better understanding which component uses which encoding and when setting an encoding is not necessary (for example, in case of using a local bus).


matthiasgeihs:
@manoranjith You suggested that the bus could also carry the initializer. Here is a proposal for that: https://github.com/perun-network/go-perun/commit/33471430d1c4ad3ea071fe14c80458208d1cd851

Let me know which approach you like better.

PR #326, opened 2022-02-16
💥 Update, ForceUpdate: allow abortion through error return value (reverts #306)

matthiasgeihs:
In #306, we disabled update abortion under the assumption that there aren't many use cases for it.

Since then, we discovered that in two of our applications we were using this functionality (perun-credential-payment and perun-websocket-backend).

Hence, I propose to enable it again.


matthiasgeihs:
@manoranjith here is one case where the error return is used

https://github.com/hyperledger-labs/perun-credential-payment/blob/98a948d52f397999b44b0b5e6418937d975efc8d/client/channel/channel.go#L156

```go
up := func(s *channel.State) error {
		// Check inputs against current state.
		curOffer, ok := s.Data.(*data.Offer)
		if !ok {
			return fmt.Errorf("data has wrong type: %T", s.Data)
		} else if !curOffer.Equal(offer) {
			return fmt.Errorf("unequal offers: got %v, expected %v", curOffer, offer)
		} else if addr := acc.Account.Address; offer.Issuer != addr {
			return fmt.Errorf("unequal addresses: got %v, expected %v", addr, offer.Issuer)
		}

		// Sign.
		sig, err := app.SignHash(acc, offer.DataHash)
		if err != nil {
			return fmt.Errorf("signing hash: %w", err)
		}

		// Update state data.
		var cert data.Cert
		copy(cert.Signature[:], sig[:])
		s.Data = &cert

		// Update balances.
		asset := s.Allocation.Assets[app.AssetIdx]
		s.Allocation.SubFromBalance(channel.Index(offer.Buyer), asset, offer.Price)
		s.Allocation.AddToBalance(ch.Idx(), asset, offer.Price)

		return nil
	}
```

We basically make sure that we are in the expected state. If not, we error and abort.
Then we create the signature. If this fails, we abort as well.

It might actually be fine here to run the checks and generate the signature before calling update.

In the case of the usage in `perun-websocket-backend`, which is a private repo, we did some decoding conversion within the update function. There it was obvious that this should have been done outside of update in the first place as it did not depend on the current state.

The thing is: If we argue that we usually don't need to check the state in the updater function, why do we use an updater function at all?

manoranjith:
@matthiasgeihs  In the example posted above, I agree that, we could do the tests outside of the updater function. Hence, the error return will not be needed.

Reg, the need for an updater function, my understanding is that the user is able to read the state of the chain and modify in place. They should be able to do this for balance and data. Validation of these modifications will be done by the `Update` function, after applying the updater.a

PR #327, opened 2022-02-22
Wire hybrid bus

RmbRT:
Enables perun-network/perun-websocket-backend#2.

### DIFF
--- a/wire/hybridbus.go
+++ b/wire/hybridbus.go
@@ -16,6 +16,7 @@ package wire
 
 import (
 	"context"
+	"fmt"
 
 	"polycry.pt/poly-go/errors"
 	"polycry.pt/poly-go/sync/atomic"
@@ -44,6 +45,12 @@ func NewHybridBus(buses ...Bus) Bus {
 	if len(buses) == 0 {
 		panic("expected at least one sub-bus.")
 	}
+	for i, bus := range buses {
+		if bus == nil {
+			panic(fmt.Sprintf("nil Bus[%d] passed to NewHybridBus", i))
+		}
+	}
+
 	return &hybridBus{buses: buses}
 }
 
--- a/wire/hybridbus.go
+++ b/wire/hybridbus.go
@@ -1,0 +1,90 @@
+// Copyright 2022 - See NOTICE file for copyright holders.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package wire
+
+import (
+	"context"
+
+	"polycry.pt/poly-go/errors"
+	"polycry.pt/poly-go/sync/atomic"
+)
+
+var _ Bus = (*hybridBus)(nil)
+
+// A Bus that can send messages over a local bus and a remote bus.
+type hybridBus struct {
+	buses []Bus
+}
+
+// NewHybridBus creates a hybrid bus that sends and receives messages over
+// multiple buses. Publishing a message over a hybrid bus will attempt to
+// publish it over all its sub-buses simultaneously. Subscribing a client to the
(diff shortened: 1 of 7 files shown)


wire/hybridbus.go
```diff
@@ -0,0 +1,90 @@
...
+	"context"
+	"fmt"
+
+	"polycry.pt/poly-go/errors"
+	"polycry.pt/poly-go/sync/atomic"
+)
+
+var _ Bus = (*hybridBus)(nil)
+
+// A Bus that can send messages over a local bus and a remote bus.
```
cryptphil, line 27:
```suggestion
// A Bus that uses multiple buses when trying to send a message.
// Could be used with a local bus and a remote bus.
```

wire/hybridbus.go
```diff
@@ -0,0 +1,90 @@
...
+func (b *hybridBus) Publish(ctx context.Context, e *Envelope) error {
+	// Whether an attempt to publish the envelope succeeded.
+	var success atomic.Bool
+	// Sub-context for sending, cancel on first success.
+	sending, cancelSending := context.WithCancel(ctx)
+	defer cancelSending()
+
+	errg := errors.NewGatherer()
+	for _, bus := range b.buses {
+		bus := bus
```
cryptphil, line 60: Maybe add a comment on why this is needed? Otherwise, this looks like a mistake.

wire/hybridbus.go
matthiasgeihs: I prefer @RmbRT 's version. It's clearer in describing what this object provides. Your suggestion focuses on how it's implemented, which is not so relevant here.

wire/hybridbus.go
```diff
@@ -0,0 +1,90 @@
...
+type hybridBus struct {
+	buses []Bus
+}
+
+// NewHybridBus creates a hybrid bus that sends and receives messages over
+// multiple buses. Publishing a message over a hybrid bus will attempt to
+// publish it over all its sub-buses simultaneously. Subscribing a client to the
+// hybrid bus will subscribe it to all its sub-buses. The sub-buses can still be
+// accessed independently for more fine-grained control.
+func NewHybridBus(buses ...Bus) Bus {
```
matthiasgeihs, line 37: This only works for a certain kind of bus, right? It should be documented here which properties a bus must fulfill in order to be used with the hybrid bus.

wire/hybridbus.go
```diff
@@ -0,0 +1,90 @@
...
+			if err == nil {
+				success.Set()
+				cancelSending()
+			}
+			return err
+		})
+	}
+
+	// Wait until all sending attempts terminated. This is when either the
+	// context expires, all buses fail, or at least one bus succeeds.
```
matthiasgeihs, line 74: This information would be valuable as part of the function documentation.

wire/hybridbus_test.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	. "perun.network/go-perun/wire"
+	"perun.network/go-perun/wire/test"
+)
+
+func TestHybridBus_New(t *testing.T) {
```
matthiasgeihs, line 26: I am not the biggest fan of such tests. They are quite tedious to maintain and don't bring much value in my opinion.

wire/net/bus_test.go
```diff
@@ -19,7 +19,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
-	"perun.network/go-perun/wire"
+	. "perun.network/go-perun/wire"
```
matthiasgeihs, line 22: We usually don't do this in go-perun. Is there a specific reason you use it here?

RmbRT: it keeps the diff small.

wire/hybridbus.go
RmbRT: Not going to comment this, as it is required and common in go. If you call an anonymous function in a goroutine inside a for loop, then all variables are referenced by name, not by value. This means that the for loop iterator variable will change before or during the execution of the goroutine. A local copy is created to pin the value the loop variable had during that iteration.
If we started commenting this, there'd be no end to the comments we would have to make.

wire/net/bus_test.go
RmbRT: This was a lazy fix, I think I had copied some code from inside the wire package, and it took less typing to fix the type names that way. I re-did it now.

wire/hybridbus.go
RmbRT: Added description to the constructor.

matthiasgeihs (approved):

PR #331, opened 2022-03-16
wallet.CloneAddress(es)

sebastianst:
- wallet: Add CloneAddress(es)
- channel: Use CloneAddresses in Params.Clone

It is convenient for users of the go-perun lib to have means for `wallet.Address` cloning. It was already done in `Params.Clone` so I extracted the functionality into the `wallet` package. No interfaces etc changed.


matthiasgeihs (approved):
👍

PR #339, opened 2022-05-06
App channel test

matthiasgeihs:
Adds an app-channel test to go-perun core, which was previously only run in the Ethereum backend.

### DIFF
--- a/client/test/progression.go
+++ b/client/test/progression.go
@@ -46,7 +46,7 @@ func makeWatcher(log log.Logger) Watcher {
 
 // HandleAdjudicatorEvent is the callback for adjudicator event handling.
 func (w *Watcher) HandleAdjudicatorEvent(e channel.AdjudicatorEvent) {
-	w.Infof("HandleAdjudicatorEvent: %v", e)
+	w.Infof("HandleAdjudicatorEvent %T: %v", e, e)
 	switch e := e.(type) {
 	case *channel.RegisteredEvent:
 		w.registered <- e
@@ -54,6 +54,8 @@ func (w *Watcher) HandleAdjudicatorEvent(e channel.AdjudicatorEvent) {
 		w.progressed <- e
 	}
 }
+
+const numStages = 3
 
 // ----------------- BEGIN PAUL -----------------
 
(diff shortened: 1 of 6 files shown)


RmbRT (dismissed):

matthiasgeihs:
Had to rebase and improve the test stability. The CI was sometimes failing as the mock backend runs very fast (basically without timeouts at the moment). Introduced a few additional stages to fix the issue.

client/test/progression.go
```diff
@@ -34,27 +34,33 @@ type Watcher struct {
...
 	progressed chan *channel.ProgressedEvent
+	concluded  chan *channel.ConcludedEvent
 }
 
 func makeWatcher(log log.Logger) Watcher {
 	return Watcher{
 		Logger:     log,
 		registered: make(chan *channel.RegisteredEvent),
 		progressed: make(chan *channel.ProgressedEvent),
+		concluded:  make(chan *channel.ConcludedEvent),
```
RmbRT, line 45: Wouldn't it be better to give those channels some buffer capacity? AFAIK you never read from `concluded`, yet you feed events into it.

matthiasgeihs: Thanks for catching this one. This was a left over from some tests I did while trying to improve stability. Indeed wasn't needed in the end.

RmbRT (approved):

client/test/progression.go
```diff
@@ -104,16 +106,17 @@ func (r *Paul) exec(_cfg ExecConfig, ch *paymentChannel) {
...
-	r.t.Logf("%v received progression confirmation 1", r.setup.Name)
-
-	// await them progressing
-	progEvent := <-r.progressed
-	r.t.Logf("%v received progression confirmation 2", r.setup.Name)
+	// Await progressed event 1.
+	r.log.Debugf("%v awaiting progressed event 1", r.setup.Name)
+	e := <-r.progressed
+	assert.Truef(e.Version() == 1, "expected version 1, got version %v", e.Version())
+	r.waitStage()
```
cryptphil, line 113: What does this `waitStage` do exactly?
It waits until a new stage has been started / the current stage has been completed s.t. we are expected to get a new progressed event?

RmbRT: It's a sync point between both test routines I afaik.

matthiasgeihs: Yes, it's an artificial synchronization point that we use in the tests. Ideally, I think, we should be able to make all tests work without those, but in some cases this would potentially require a rework of some architecture parts.

PR #341, opened 2022-05-19
Client tests: fail on error

matthiasgeihs:
Client tests involve concurrency. t.FailNow is not available there.
That's why we often used assert instead of require. Now we can use
role.Require.

### DIFF
--- a/client/test/channel.go
+++ b/client/test/channel.go
@@ -16,11 +16,10 @@ package test
 
 import (
 	"context"
+	"errors"
 	"io"
 	"math/big"
 	"time"
-
-	"github.com/stretchr/testify/assert"
 
 	"perun.network/go-perun/channel"
 	"perun.network/go-perun/client"
@@ -73,7 +72,7 @@ func (ch *paymentChannel) openSubChannel(
 
 	prop := ch.r.SubChannelProposal(rng, cfg, ch.Channel, &initAlloc, app)
 	subchannel, err := ch.r.ProposeChannel(prop)
-	assert.NoError(ch.r.t, err)
+	ch.r.RequireNoError(err)
 	ch.r.log.Infof("New subchannel opened: %v", subchannel.Channel)
 
 	for i, bal := range initBals {
@@ -88,7 +87,7 @@ func (ch *paymentChannel) acceptSubchannel(
 	initBals []*big.Int,
 ) *paymentChannel {
 	subchannel, err := propHandler.Next()
-	assert.NoError(ch.r.t, err)
+	ch.r.RequireNoError(err)
 	ch.r.log.Infof("New subchannel opened: %v", subchannel.Channel)
 
 	for i, bal := range initBals {
@@ -105,7 +104,7 @@ func (ch *paymentChannel) sendUpdate(updater func(*channel.State), desc string) 
 
 	err := ch.Update(ctx, updater)
 	ch.log.Infof("Sent update: %s, err: %v", desc, err)
-	assert.NoError(ch.r.t, err)
+	ch.r.RequireNoError(err)
 }
 
 func (ch *paymentChannel) sendTransfer(amount channel.Bal, desc string) {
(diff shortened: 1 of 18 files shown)


client/test/role.go
```diff
@@ -122,15 +125,20 @@ func ExecuteTwoPartyTest(ctx context.Context, role [2]Executer, cfg ExecConfig)
...
 		}(i)
 	}
 
 	// wait for clients to finish or timeout
-	wg.WaitCtx(ctx)
-	return ctx.Err()
+	select {
+	case <-wg.WaitCh():
+	case <-ctx.Done():
+		t.Fatal(ctx.Err())
```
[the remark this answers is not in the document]
cryptphil: Another approach would be to call `runtime.GoExit()` inside of `RequireNoError()` etc. which will then end the current go-routine.

client/test/role.go
```diff
@@ -284,6 +292,43 @@ func (r *role) ProposeChannel(req client.ChannelProposal) (*paymentChannel, erro
 	return ch, nil
 }
 
+func (r *role) RequireNoError(err error) {
```
cryptphil, line 295: As written above, it would be possible to call `runtime.GoExit()` in those `Require` methods, which will lead to stopping the go-routine in which they are called.

client/test/subchannel_dispute.go
```diff
@@ -114,7 +111,7 @@ func (r *DisputeSusie) exec(_cfg ExecConfig, ledgerChannel *paymentChannel) {
 	m := channel.MakeStateMap()
 	m.Add(subState0.State)
 	err = r.setup.Adjudicator.Withdraw(ctx, reqLedger, m)
-	assert.Error(err, "withdraw should fail because other party should have refuted.")
+	r.RequireTruef(err != nil, "withdraw should fail because other party should have refuted.")
```
cryptphil, line 114: Why not use `r.RequireNoErrorf()` here?

client/test/channel.go
```diff
@@ -150,22 +150,21 @@ func (ch *paymentChannel) assertBals(state *channel.State) {
 		ch.bals[0], ch.bals[1],
 		bals[0], bals[1],
 	)
-	assert := assert.New(ch.r.t)
-	assert.Zerof(bals[0].Cmp(ch.bals[0]), "bal[0]: %v != %v", bals[0], ch.bals[0])
-	assert.Zerof(bals[1].Cmp(ch.bals[1]), "bal[1]: %v != %v", bals[1], ch.bals[1])
+	ch.r.RequireTruef(bals[0].Cmp(ch.bals[0]) == 0, "bal[0]: %v != %v", bals[0], ch.bals[0])
```
cryptphil, line 153: The function is still called `assertBals` even though we are only using `require` instead of `assert` here.

client/test/mallory.go
```diff
@@ -73,35 +70,35 @@ func (r *Mallory) exec(_cfg ExecConfig, ch *paymentChannel) {
...
 	defer waitCancel()
 	// refutation increased the timeout.
-	assert.NoError(event.Timeout().Wait(waitCtx))
+	r.RequireNoError(event.Timeout().Wait(waitCtx))
 
 	wdCtx, wdCancel := context.WithTimeout(context.Background(), r.timeout)
 	defer wdCancel()
 	err = r.setup.Adjudicator.Withdraw(wdCtx, req0, nil)
-	assert.Error(err, "withdrawing should fail because Carol should have refuted.")
+	r.RequireTruef(err != nil, "withdrawing should fail because Carol should have refuted")
```
cryptphil, line 101: Why not use `r.RequireNoErrorf()` here?

cryptphil (changes requested):

client/test/role.go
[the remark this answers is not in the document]
matthiasgeihs: Thx for checking. In fact, with the previous version, the same problem exists. So we don't solve this here but at least we can let the test quickly know that something is going wrong. To be honest, it also doesn't seem super problematic that we don't shutdown the go routines as this happens automatically when the test fails. In fact, it might be problematic to call `runtime.GoExit()` (I was considering that), because then you might accidentally shutdown the main routine if the tests are not setup as expected.

client/test/channel.go
matthiasgeihs: I think it is fine. Asserting can also mean that you don't continue afterwards.

client/test/mallory.go
matthiasgeihs: Because we require an error.

client/test/role.go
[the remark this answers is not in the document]
cryptphil: Yes, I also saw that in principle the same problem existed before and is also the reason why using `require` does not work in go-routines since it also just calls `t.FailNow()`.

And true, using the approach with `runtime.GoExit()` may cause unwanted behavior when using it without caution.

client/test/channel.go
cryptphil: Hm okay.

client/test/mallory.go
cryptphil: My bad, you're correct.

PR #342, opened 2022-05-19
Satisfy linter

matthiasgeihs:
My local version of golangci for some reason reports more than our CI. Goal of this PR: Satisfy local and remote golangci.

depends on #341

### DIFF
--- a/channel/mock_app.go
+++ b/channel/mock_app.go
@@ -116,17 +116,29 @@ func (a MockApp) NewData() Data {
 
 // ValidTransition checks the transition for validity.
 func (a MockApp) ValidTransition(params *Params, from, to *State, actor Index) error {
-	return a.execMockOp(from.Data.(*MockOp))
+	op, ok := from.Data.(*MockOp)
+	if !ok {
+		return fmt.Errorf("wrong data type: expected *MockOp, got %T", from.Data)
+	}
+	return a.execMockOp(op)
 }
 
 // ValidInit checks the initial state for validity.
 func (a MockApp) ValidInit(params *Params, state *State) error {
-	return a.execMockOp(state.Data.(*MockOp))
+	op, ok := state.Data.(*MockOp)
+	if !ok {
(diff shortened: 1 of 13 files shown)


channel/mock_app.go
```diff
@@ -116,30 +116,52 @@ func (a MockApp) NewData() Data {
 
 // ValidTransition checks the transition for validity.
 func (a MockApp) ValidTransition(params *Params, from, to *State, actor Index) error {
-	return a.execMockOp(from.Data.(*MockOp))
+	op, ok := from.Data.(*MockOp)
```
cryptphil, line 119: Maybe it would make sense to create a function like the one below so that you don't have to replicate these 4 lines in all those methods. 

```go
func getMockOp(data Data) (*MockOp, error) {
	if op, ok := data.(*MockOp); ok {
		return op, nil
	}
        return nil, fmt.Errorf("wrong data type: expected *MockOp, got %T", data)
}

cryptphil (commented):
LGTM, just one small comment

channel/mock_app.go
cryptphil: It seems that `MockOp` is also converted from an `Action` type, so one would need to change the parameter type to `interface{}`.

matthiasgeihs: You are right, we could reduce code duplication, but at the expense of clarity. Also the code doesn't actually become shorter, it becomes longer because we still need to check for the error. We could create a custom error type so that we don't repeat the error message every time, but we don't usually do this in go-perun, so I didn't do it here either. In Go it can be idiomatic to rather copy small parts of code instead of creating additional dependencies.


Btw: If you only have minor optional comments, you can approve the PR and leave it up to the author to address them.

cryptphil: Alright (and yes, my bad, I should have already approved it).

cryptphil (approved):

PR #343, opened 2022-05-19
Simplify virtual channel proposal

matthiasgeihs:
Before, it was quite involved to generate a virtual channel proposal. One had to provide so called index maps that specify how participant indices are remapped from root channel to virtual channel. For our current case of two-party, it is, however, quite simple to generate these based with a more simpler input (a boolean flag that denotes whether the participant order is the same or different).

depends on #342


matthiasgeihs:
Closed due to inactivity.

PR #349, opened 2022-06-02
Settle channel when funding fails (Variant B)

cryptphil:
This is the second implementation variant where the `ProposeChannel` method does not automatically settle the channel after a failed funding but the client itself has to take care of this.

### DIFF
--- a/channel/funder.go
+++ b/channel/funder.go
@@ -31,10 +31,9 @@ type (
 		// because a peer did not fund the channel in time.
 		// Depending on the funding protocol, if we fund first and then the peer does
 		// not fund in time, a dispute process needs to be initiated to get back the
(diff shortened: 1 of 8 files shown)


cryptphil:
Regarding testing the specific error type when the channel creation fails: we could only check whether the error for the party with the non-failing funder is of type `FundingTimeoutError`, however, this would make the tests slightly more complicated and I don't think this is really worth it.

matthiasgeihs (commented):
It is essential to return a specific error type. Only then can we check whether it is actually worth trying to settle. (If we failed before trying to fund there is no reason for trying to settle.)

I added commits to adapt the functionality and revised documentation. Please have a look.

matthiasgeihs (approved):

PR #351, opened 2022-06-24
Remove settle secondary

### LINKED ISSUE #81: Move settle secondary logic to user, remove `secondary` parameter
**Context:** Parameter `secondary bool` of function `Channel.Settle` is part of a protocol optimization that aims to reduce client gas fees. The Ethereum contracts require only one client to call `conclude`. The other clients can skip the `conclude` call and thereby save some gas fees.

**Problem:** I would like to discuss the following issues with the current implementation:
1. The `secondary` parameter is not document well and it is not clear to a new framework user how to use it.
2. It is arguable whether the complexity introduced by the settle secondary logic outweighs its benefits.
3. It is arguable whether implementing the settle secondary logic on the framework level is preferable over implementing it on the application level.

**Suggestion:**
I see the following options to improve on the situation:
- **Keep the logic:** Make parameter `secondary` optional and document it well.
- **Drop the logic:** If required, implement the secondary logic on the application level.

matthiasgeihs:
This PR removes the settle secondary logic. It was not well documented and rarely used, and lead to confusion. We can still implement the same functionality on the application level if necessary.

Closes #81

### DIFF
--- a/channel/adjudicator.go
+++ b/channel/adjudicator.go
@@ -103,17 +103,11 @@ type (
 
 	// An AdjudicatorReq collects all necessary information to make calls to the
 	// adjudicator.
-	//
-	// If the Secondary flag is set to true, it is assumed that this is an
-	// on-chain request that is executed by the other channel participants as well
-	// and the Adjudicator backend may run an optimized on-chain transaction
-	// protocol, possibly saving unnecessary double sending of transactions.
 	AdjudicatorReq struct {
-		Params    *Params
-		Acc       wallet.Account
-		Tx        Transaction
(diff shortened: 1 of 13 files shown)


cryptphil (dismissed):

matthiasgeihs:
Wait with merging until finished multi-ledger integration in perun-eth-backend.

cryptphil (approved):

matthiasgeihs:
> If we want to get this behavior back, we could still do so on the user side by just waiting for some time if we get the desired Adjudicator event, right?

Yes.

PR #358, opened 2022-07-14
readme fix of pdf link to paper

dzmitry-lahoda:
Signed-off-by: Dzmitry Lahoda <dzmitry@lahoda.pro>


matthiasgeihs:
Hi, thx for the fix. Currently, it is requirement that all committers are listed in the NOTICE file. However, I will discuss with the other maintainers if we want to keep this requirement and come back to you.

matthiasgeihs:
Hey @dzmitry-lahoda, I checked with the other maintainers. Would you mind adding yourself to the NOTICE file?

matthiasgeihs (commented):
Looks good, thanks for the contribution. You only need to change the order of the commits. The addition to the NOTICE file must come before the first commit.

matthiasgeihs:
@dzmitry-lahoda, I just double checked why our copyright script is complaining. It seems like you are using a different E-Mail address in your commits than the one you included in the NOTICE file. You need to: either change the address used in the commits or in the NOTICE file. You can also run `.scripts/check-notice-authors.sh` locally to see if that works.

Sorry for the inconvenience.

matthiasgeihs:
Closed due to inactivity, addressed in #379.

PR #360, opened 2022-07-19
Fix local watcher deadlock

matthiasgeihs:
In rare occasions, the local watcher was running into a deadlock when stopping and handling an event at the same time.


manoranjith:
This is really a tricky flaw, that is also hard to consistently reproduce (because it occurs only under certain timing conditions).

Good catch. The fix LGTM.

manoranjith (approved):

PR #363, opened 2022-07-25
MockBackend: Remove context from subscription

matthiasgeihs:
This PR includes a small update to the mock subscription. Storing a context in a struct is disadvised and not needed here as we can just rely on the functionality of the events channel.


manoranjith (approved):
LGTM

PR #366, opened 2022-08-01
Fix Multi-Adjudicator Subscription

### LINKED ISSUE #365: Cannot close multi-ledger Channel
### Location
`multi/`
To reconstruct the bug: Add `chBobAlice.Close()` to `client/test/multiledger_dispute.go:135` and then run the multi-ledger dispute tests.

### Problem
When trying to close a multi-ledger channel with `ch.Close()` the method does not return and we get a deadlock.
This only happens if we have started a watcher for this channel.

### Proposal
We have to identify the reason for this bug and then fix it.

cryptphil:
Closes #365.


matthiasgeihs (approved):
I added a commit that ensures that the for loop will return when the subscription is closed. Please have a look. Otherwise looks good.

PR #369, opened 2022-08-02
Backend timeout

### LINKED ISSUE #368: Implement timeouts for mock backend subscription
Side node: As the MockBackend used for testing the core does not use proper time out logic yet, these things are easily missed in the core client tests. We should implement the timeout logic for events emitted by the Mock Backend Subscription.

_Originally posted by @matthiasgeihs in https://github.com/hyperledger-labs/go-perun/issues/367#issuecomment-1203132061_

matthiasgeihs:
Closes #368


client/test/backend.go
```diff
@@ -175,11 +176,12 @@ func (b *MockBackend) Progress(_ context.Context, req channel.ProgressReq) error
 	b.mu.Lock()
 	defer b.mu.Unlock()
 
+	timeout := time.Now().Add(time.Duration(req.Params.ChallengeDuration) * time.Millisecond)
```
cryptphil, line 179: Shouldn't the ChallengeDuration be given in seconds?

cryptphil (changes requested):
Looks good, just a small remark.

client/test/backend.go
matthiasgeihs: Yes, but this would be too long for the tests. Here we scale down to milliseconds.

cryptphil: Okay, makes sense.

cryptphil (approved):

PR #370, opened 2022-08-02
Recover fund abstract

matthiasgeihs:
We need the fund recovery test to be abstract so that it can be reused in backend implementations.

Relates to https://github.com/hyperledger-labs/perun-eth-backend/issues/20

Based on #369


client/test/fund.go
```diff
@@ -0,0 +1,135 @@
...
+// FundSetup represents the fund recovery test parameters.
+type FundSetup struct {
+	ChallengeDuration uint64
+	FridaInitBal      channel.Bal
+	FredInitBal       channel.Bal
+}
+
+// TestFundRecovery performs a test of the fund recovery functionality for the
+// given setup.
+func TestFundRecovery(ctx context.Context, t *testing.T, params FundSetup, setupRoles func(*rand.Rand) [2]RoleSetup) { //nolint:revive // test.Test... stutters but this is OK in this special case.
```
cryptphil, line 42: Why not just name it `FundRecovery`?
Calling it as `test.FundRecovery` also sounds fine IMO.

cryptphil (dismissed):
Looks good to me, I just added a comment regarding the name of the test function. :smile:

client/test/fund.go
matthiasgeihs: Then it sounds like a data structure. Since it's a test function, it should also read like it runs a test.

client/test/backend.go
```diff
@@ -250,7 +288,7 @@ func (b *MockBackend) checkState(s *channel.State, op checkStateFunc) error {
 }
 
 // Withdraw withdraws the channel funds.
-func (b *MockBackend) Withdraw(_ context.Context, req channel.AdjudicatorReq, subStates channel.StateMap) error {
+func (b *MockBackend) Withdraw(_ context.Context, req channel.AdjudicatorReq, subStates channel.StateMap, acc wallet.Address) error {
```
cryptphil, line 291: Just for my comprehension: This change is necessary to have similar behavior as with the Ethereum backend when using the adjudicator's receiver address to change the address the funds are sent to, correct?

client/test/fund.go
```diff
@@ -0,0 +1,160 @@
...
+type FundSetup struct {
+	ChallengeDuration uint64
+	FridaInitBal      channel.Bal
+	FredInitBal       channel.Bal
+	BalanceDelta      channel.Bal
+}
+
+// TestFundRecovery performs a test of the fund recovery functionality for the
+// given setup.
+func TestFundRecovery(ctx context.Context, t *testing.T, params FundSetup, setup func(*rand.Rand) ([2]RoleSetup, channel.Asset)) { //nolint:revive // test.Test... stutters but this is OK in this special case.
```
cryptphil, line 42: Since the method signature is quite long I would probably wrap it at some point but that's up to you.

client/test/fund.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Store client balances before running test.
+	balancesBefore := channel.Balances{
+		{
+			frida.BalanceReader.Balance(asset),
+			fred.BalanceReader.Balance(asset),
+		},
+	}
+
+	// The channel into which Fred's created ledger channel is sent into.
```
cryptphil, line 101:
```suggestion
	// The channel into which Fred's created ledger channel is sent.
```
I think the mistake even comes from me when I originally created the tests but I didn't see it until now.

cryptphil (dismissed):

client/test/backend.go
matthiasgeihs: Exactly. In fact, it was also necessary to adapt the Balance Reader interface, because of that.

client/test/fund.go
matthiasgeihs: Done

client/test/fund.go
matthiasgeihs: Revised the comment.

cryptphil (approved):

PR #373, opened 2022-08-19
Allow retry Channel.ForceUpdate

matthiasgeihs:
If a ForceUpdate fails, the channel may remain in state `Progressing`.
With this change we allow for retrying from this phase.

### DIFF
--- a/channel/machine.go
+++ b/channel/machine.go
@@ -401,7 +401,7 @@ func (m *machine) SetRegistered() error {
 // the given state.
 func (m *machine) SetProgressing(state *State) error {
 	if !inPhase(m.phase, []Phase{Registered, Progressing, Progressed}) {
-		return m.phaseErrorf(m.selfTransition(), "can only progress when registered or progressed")
+		return m.phaseErrorf(m.selfTransition(), "can only progress after registration")
 	}
 	m.setStaging(Progressing, state)
 	return nil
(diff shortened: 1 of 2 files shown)


channel/machine.go
```diff
@@ -400,7 +400,7 @@ func (m *machine) SetRegistered() error {
 // SetProgressing sets the machine phase to Progressing and the staging state to
 // the given state.
 func (m *machine) SetProgressing(state *State) error {
-	if !inPhase(m.phase, []Phase{Registered, Progressed}) {
+	if !inPhase(m.phase, []Phase{Registered, Progressing, Progressed}) {
 		return m.phaseErrorf(m.selfTransition(), "can only progress when registered or progressed")
```
cryptphil, line 404:
```suggestion
		return m.phaseErrorf(m.selfTransition(), "can only progress when registered, progressing, or progressed")
```

client/adjudicate.go
```diff
@@ -447,7 +447,9 @@ func (c *Channel) subChannelStateMap() (states channel.StateMap, err error) {
 // ensureRegistered ensures that the channel is registered.
 func (c *Channel) ensureRegistered(ctx context.Context) error {
 	phase := c.Phase()
-	if phase == channel.Registered || phase == channel.Progressed {
+	if phase == channel.Registered ||
+		phase == channel.Progressing ||
```
cryptphil, line 451: Why do we also treat a channel as registered if it's currently still progressing?

cryptphil: Answered in a meeting.

cryptphil (dismissed):

channel/machine.go
matthiasgeihs: revised error message

PR #378, opened 2022-10-05
Create app id type

### LINKED ISSUE #352: Separate type for channel.AppID
### Location
`channel/app.go`

### Problem
Currently, an app identifier is just a wallet address. This may make sense for Ethereum smart contracts, but does not generalize well to other scenarios where the app identifier may be an application defined string, for example.

### Proposal
Create an abstract type `channel.AppID` that allows for application specific implementation of the app identifier.

matthiasgeihs:
Closes #352.

### DIFF
--- a/apps/payment/app.go
+++ b/apps/payment/app.go
@@ -24,12 +24,12 @@ import (
 
 // App is a payment app.
 type App struct {
-	Addr channel.AppID
+	ID channel.AppID
 }
 
 // Def returns the address of this payment app.
 func (a *App) Def() channel.AppID {
-	return a.Addr
+	return a.ID
 }
 
 // NewData returns a new instance of data specific to the payment app,
--- a/apps/payment/app.go
+++ b/apps/payment/app.go
@@ -20,16 +20,15 @@ import (
 
 	"perun.network/go-perun/channel"
 	"perun.network/go-perun/log"
-	"perun.network/go-perun/wallet"
 )
 
 // App is a payment app.
 type App struct {
-	Addr wallet.Address
+	Addr channel.AppID
 }
 
 // Def returns the address of this payment app.
-func (a *App) Def() wallet.Address {
+func (a *App) Def() channel.AppID {
 	return a.Addr
 }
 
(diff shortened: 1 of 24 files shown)


apps/payment/app.go
```diff
@@ -20,16 +20,15 @@ import (
 
 	"perun.network/go-perun/channel"
 	"perun.network/go-perun/log"
-	"perun.network/go-perun/wallet"
 )
 
 // App is a payment app.
 type App struct {
-	Addr wallet.Address
+	Addr channel.AppID
```
cryptphil, line 27: I think it would make more sense to then also rename it from `Addr` to ID`.

backend/sim/channel/app.go
```diff
@@ -0,0 +1,52 @@
...
+	"perun.network/go-perun/backend/sim/wallet"
+	"perun.network/go-perun/channel"
+)
+
+// AppID represents an app identifier.
+type AppID struct {
+	*wallet.Address
+}
+
+// Equal returns whether the object is equal to the given object.
```
cryptphil, line 29: Isn't this doc a bit too general regarding "object" instead of AppID?

channel/appregistry_internal_test.go
```diff
@@ -102,13 +102,50 @@ func assertIdentity(t *testing.T, expected App) {
...
+	copy(id[:], data)
+	return nil
+}
+
+func (id appID) Equal(b AppID) bool {
+	bTyped, ok := b.(appID)
+	return ok && bytes.Equal(id[:], bTyped[:])
+}
+
+// Key returns the object key which can be used as a map key.
```
cryptphil, line 144: You could remove this doc since the other functions also have no doc and its explained in the interface imo.

cryptphil (dismissed):
Just left some comments about the doc, the rest looks good to me.

apps/payment/app.go
matthiasgeihs: Done

backend/sim/channel/app.go
matthiasgeihs: I don't see a problem with being general here. It's still clear, no?

channel/appregistry_internal_test.go
matthiasgeihs: Removed

matthiasgeihs:
Addressed review comments and rebased. @cryptphil

PR #397, opened 2024-02-15
Egoistic funding

sophia1ch:
## Description:
This PR builds upon the changes introduced in https://github.com/hyperledger-labs/perun-eth-backend/pull/45 by extending egoistic funding to cross-chain swaps in the eth-backend.

In this update, we introduce a map for each chain, designating whether a chain is considered egoistic or not. This distinction is crucial in determining the funding order for different chains, ensuring that egoistic chains are funded last.


client/proposal.go
```diff
@@ -127,6 +127,18 @@ func (r *ProposalResponder) Accept(ctx context.Context, acc ChannelProposalAccep
 	return r.client.handleChannelProposalAcc(ctx, r.peer, r.req, acc)
 }
 
+// SetEgoisticChain sets the egoistic chain flag for a given ledger.
+func (r *ProposalResponder) SetEgoisticChain(egoistic multi.LedgerID) {
```
cryptphil, line 131: So you can only set the egoistic flag when you receive a proposal? Can't you set this in beforehand when initializing the funders? In the case of Perun-X, the hub would just set this for all chains?

cryptphil (changes requested):

NhoxxKienn:
I think the protocol is to merge first to our fork before create a PR here, right?

cryptphil:
> I think the protocol is to merge first to our fork before create a PR here, right?

No, in general, you always merge to the hyperledger org main and then only sync this fork on perun-network.

client/proposal.go
cryptphil: Hm okay I see, thanks.

cryptphil: Hmm okay, I understand, thanks.

MAINTAINERS.md
```diff
@@ -1,16 +1,16 @@
...
 ## Active Maintainers
-| Name              | Github    | [Discord][_chat_url]        |
-|-------------------|-----------|----------------|
-| Hendrik Amler     | [@tinnendo](https://github.com/tinnendo) | hendrik#5345 |
-| Jan Bormet    | [@janbormet](https://github.com/janbormet) | _.pants |
-| Philipp-Florens Lehwalder | [@cryptphil](https://github.com/cryptphil) | cryptphil |
-| Steffen Rattay  | [@rmbrt](https://github.com/rmbrt) | rmbrt |
-| Ilja von Hoessle  | [@iljabvh](https://github.com/iljabvh) | iljabvh |
+| Name              | Github                                             | [Discord][_chat_url]        |
+|-------------------|----------------------------------------------------|----------------|
```
cryptphil, line 5: Minor: The authors are usually sorted alphabetically by their last name

cryptphil (approved):

Issue #395, opened 2024-01-19
Failing unit tests


janbormet:
### Location
`TestAppRandomizerSet` in `channel/test/app_randomizer_internal_test.go`
and
`TestDialer_Dial`, happy case in `wire/net/simple/dialer_internal_test.go`

### Problem
There seem to be two failing unit tests. Apparently they do not fail in the CI nor for some of the maintainers, so there might be a race? 
These are the error traces (from `go test ./... --count=1` in content root):

#### TestAppRandomizerSet
```
pkg/test: using rootSeed 1705672702804719252
--- FAIL: TestAppRandomizerSet (0.00s)
    app_randomizer_internal_test.go:35: 
                Error Trace:    app_randomizer_internal_test.go:35
                Error:          Should be false
                Test:           TestAppRandomizerSet
                Messages:       appRandomizer should have changed
FAIL
FAIL    perun.network/go-perun/channel/test     0.004s
```

#### TestDialer_Dial
```
pkg/test: using rootSeed 1705672703035073752
--- FAIL: TestDialer_Dial (0.00s)
    --- FAIL: TestDialer_Dial/happy (0.00s)
        dialer_internal_test.go:88: 
                Error Trace:    dialer_internal_test.go:88
                                                        concurrent.go:223
                                                        goexit.go:90
                                                        goexit.go:92
                                                        asm_amd64.s:1650
                Error:          Not equal: 
                                expected: &wire.Envelope{Sender:(*simple.Address)(0xc000206030), Recipient:(*simple.Address)(0xc000206040), Msg:(*wire.PingMsg)(0xc00021c018)}
                                actual  : &wire.Envelope{Sender:(*wire.Address)(0xc000022480), Recipient:(*wire.Address)(0xc0000223e0), Msg:(*wire.PingMsg)(0xc0000125e8)}
                                
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,4 +1,10 @@
                                 (*wire.Envelope)({
                                - Sender: (*simple.Address)((len=32) "\x98\xcf\\\xa4{\xd6x\xceS\x18\x1d\xe8>E\x91\xfcJȞDv\xf6\x8f\xad\x9a\xbaSՂӿ\x93"),
[...]

PR #399, opened 2024-02-19
#395_Fix_Falling_Unit_Tests

NhoxxKienn:
## Description
This PR serves to fix the #395 issue of failing unit tests. 

### App Randomizer Internal Test 
Location: `channel/test/app_randomizer_internal_test.go`

Cause: The issue stems from the lack of Comparator (`Equal`) for MockAppRandomizer. This leads to the test not recognizing that the default appRandomizer has already been set in the internal test.

Solution: Add an identifier to the MockAppRandomizer to compare the old and new appRandomizer.

### Dialer Internal Test
Location: (`wire/net/simple/dialer_internal_test.go`)

Cause: Like @RmbRT has pointed out, there is a race between the init of `wire/simple` and `backend/sim/wire`, which set the RandomAddress and RandomAccount functions to be used in the internal tests.

Solution: Add the initialization of `NewRandomAddress` and `NewRandomAccount` to use the implementation from the `wire/simple`.

Alternative: Remove `init()` of `wire/simple` and only use the init of `backend/sim/wire`. This also results in passing internal tests but might cause issues for packages using `wire/simple` (need further investigation).

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -9,6 +9,7 @@
 | Steffen Rattay  | [@rmbrt](https://github.com/rmbrt) | rmbrt |
 | Ilja von Hoessle  | [@iljabvh](https://github.com/iljabvh) | iljabvh |
 | Jens Winkle       | [@DragonDev1906](https://github.com/DragonDev1906) | jens#4601 |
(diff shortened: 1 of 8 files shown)


iljabvh (approved):
LGTM

Issue #421, opened 2025-05-16
Net Bus Exchange Address Bug


NhoxxKienn:
### Location

`wire/net/account.go` in `NewAuthResponseMsg` at line `101`

### Problem

Exchange Address in `wire/net/bus` requires a handshake protocol to validate the authenticity of the peer`s user account.

The current implementation `go-perun v0.13.0` extracts the account from the fixed index of `testBackendID`. This will only work in a test environment and wont work with other backends.

### Proposal

Adjust the implementation so that the AuthResponseMsg will extract the wire Account from a specific backend instead of hard-coding it now.

NhoxxKienn:
solved with PR #422

PR #422, opened 2025-05-19
fix(wire/net): remove test backend signature from authMsg of exchange address protocol

NhoxxKienn:
This PR fix the bug #421 by removing the hard-coded test-backend signature from AuthMsg and required the user to sign all of the address map using all of there backends.

## Location
`wire/net` package 

## Disclaimer
This implementation increase the number of authentication msgs by the number of backends (log(n)) and could cause performance issues with a large number of backends.

Signed-off-by: Minh Huy Tran <huy@perun.network>

### DIFF
--- a/wire/account.go
+++ b/wire/account.go
@@ -40,8 +40,6 @@ type Account interface {
 	Sign(msg []byte) ([]byte, error)
 }
 
-const testBackendID = 0
-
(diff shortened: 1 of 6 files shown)


sophia1ch (approved):
LGTM

